PR #12, opened 2021-04-02
solution added

TimurHrynenko:
- [DEMO LINK](https://TimurHrynenko.github.io/layout_calendar/)
- [TEST REPORT LINK](https://TimurHrynenko.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,50 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  max-width: getCalendarWidth( $daySize, $dayGap, $calendarPadding );
+  padding: $calendarPadding;
+  margin: auto;
+  gap: $dayGap;
+
+  @each $day, $order in $week {
+    &_start-day_#{$day} {
+      :first-child {
+        margin-left: getFirstDayShift( $daySize, $dayGap, $order );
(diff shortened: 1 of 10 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+  }
+
+  &__day {
+    @include daySize($daySize);
+    background-color: #eee;
+    border: 1px solid black;
+    text-align: center;
+    transition: transform 0.5s ease, background-color 0.5s ease;
```
rauzok, line 22:
```suggestion
    transition-duration: 0.5s;
```
You can use simply that too, and it works.

rauzok (approved):
Great work!

src/styles/blocks/calendar.scss
TimurHrynenko: Yep, but, still, I think its simplier to use combined property `transition`, than separate properties like `transition-duration` and so on, because in most common case instead of specific values like `transform `or `background-color` , there will be a value `all`. 
But your suggestion was interesting, thank you.

aionova: You're absolutely right, Timur

aionova (approved):
Good job!

PR #18, opened 2021-04-03
solution

KapToH337:
[DEMO LINK](https://KapToH337.github.io/layout_calendar/)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar__month-length--29">
+      <div class="calendar__day calendar__day--1 calendar__day-start--sun"></div>
```
rauzok:
```suggestion
    <div class="calendar calendar__month-length--29 calendar__day-start--sun">
      <div class="calendar__day calendar__day--1 "></div>
```
It has to be together in block

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+      content: "#{$number}";
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      position: relative;
+      top: 33%;
+    }
+  }
+
+  &__day {
+    margin: 0 0 1px 1px;
```
rauzok:
```suggestion
    display: flex;
    gap: 1px;
```
Use **gap** instead margin

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+
+@keyframes slidein {
+  from {
+    transform: translate(0, 0);
+  }
+
+  to {
+    transform: translate(0, -20px);
+  }
+}
```
rauzok: When I hover the mouse cursor over the block there is an animation, when I remove the mouse cursor from the block there is no reverse animation.
I think you should fix it.

rauzok (changes requested):
Good job, but you should fix something.

rauzok (approved):
Cool work!

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
   margin: 0;
 }
+
+.calendar {
+  margin: 10px 10px 9px;
+  width: $calendar-size;
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: row;
+
```
rauzok, line 15:
```suggestion
    gap: 1px;
    
```

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+  @for $index from 1 to 31 {
+    &__month-length--#{$index} :nth-child(n + #{$index+1}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    display: flex;
+    gap: 1px;
```
rauzok:
```suggestion
```

rauzok (changes requested):

rauzok (approved):
Good!

aionova (changes requested):
Please re-deploy it again, as last changes weren't applied

src/index.html
```diff
@@ -7,6 +7,38 @@
...
+      <div class="calendar__day calendar__day--2"></div>
```
aionova, line 13: What if we remove classes from each div? Can we do it without them?

aionova (changes requested):

src/index.html
KapToH337: I think it's impossible, because this classes need for @for and @each loop

KapToH337:
now worked or re-deploy?
I don't understand what changes are not applied

aionova (approved):
Good job!

PR #27, opened 2021-04-04
Added content

chumasov:
[DEMO LINK](https://chumasov.github.io/layout_calendar/)
[TEST REPORT LINK](https://chumasov.github.io/layout_calendar/report/html_report/)


aionova (changes requested):
Please add scss files and make one more commit with them

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+    background-color: $dayColor;
+    border: 1px solid #000;
+    font-family: Arial, sans-serif;
+    font-size: 30px;
+
+    &:hover {
+      cursor: pointer;
+      background-color: $hoverColor;
+      transform: translate(0, -20px);
+      transition: 0.5s;
```
aionova, line 34: indicate transition in &__day, not in hover. Then transition will work in both ways (when block with day goes up and down)

aionova (approved):
Good job!
p.s. No need to include css files, scss will be enough)

PR #32, opened 2021-04-05
layout_calendar

dmytro-abram:
[DEMO LINK](https://dmytro-abram.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="content">
+    <main class="calendar calendar--start-day--thu calendar--month-length--30">
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -6,7 +6,39 @@
...
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
```
leonid-vegera, line 41: empty divs... I suppose it is OK, but it would be better to add any class for them, for future

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,49 @@
+.content {
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
```
leonid-vegera, line 6: It seems to me Misha told us to create a separate folder for every block every time

src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,3 @@
+body {
+  margin: 0;
+}
```
leonid-vegera, line 3: tag body pointed two times. fonts have to be indicated better in html tag

leonid-vegera (approved):
Good job!

aionova (approved):
Great!

PR #57, opened 2021-04-06
first ver

cheytac200:


aionova (changes requested):
Please add demo-link

cheytac200:
- [DEMO LINK](https://cheytac200.github.io/layout_calendar/)

src/styles/calendar.scss
```diff
@@ -0,0 +1,26 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  grid-gap: 1px;
```
aionova:
```suggestion
  gap: 1px;
```
It's not grid

src/styles/calendar.scss
```diff
@@ -0,0 +1,26 @@
...
+    background-color: #eee;
+    border: 1px solid #000;
+
+  }
+
+  &__day:hover {
+    background-color: #ffbfcb;
+    cursor: pointer;
+    transform: translateY(-20px);
+    transition: 0.5s;
```
aionova: move transition to  &__day selector, and then each block will move smoothly in both sides, not only when the block 'goes up', but down too. Also please include which exact property you want to use transition for (transform, background-color)

aionova (changes requested):

src/styles/calendar.scss
cheytac200: @aionova "Also please include which exact property you want to use transition for (transform, background-color)" не могу понять что нужно

aionova (approved):
Great!

PR #189, opened 2021-07-13
first commit

VVHimiak:
- [DEMO LINK](https://VVHimiak.github.io/layout_calendar/)
- [TEST REPORT LINK](https://VVHimiak.github.io/layout_calendar/report/html_report/)


DmytroMykoliv:
В тебе стилі не загрузились, зроби "git add ." , потім перевір статус git status , якщо все готове(зелене) то коміт і пуш

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar-wrapper">
+      <main class="calendar calendar--month-length--29">
+        <div class="calendar__day calendar__day--1 calendar__day--start-day-sunday"></div>
```
ihor-jpeg: The calendar__day--start-day-$$$ modifier should be applied to the calendar container, not to the first element. 
In your case it should be applied to the "main" tag.

src/index.html
```diff
@@ -7,6 +7,40 @@
...
+        <div class="calendar__day calendar__day--2"></div>
```
ihor-jpeg: You don't need calendar__day--$ modifier. Reference each day through css `:nth-child property`

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
-body {
-  margin: 0;
-}
+@import "./utilities/variables";
```
ihor-jpeg, line 1: I can't see your individual css files

ihor-jpeg (changes requested):

src/styles/blocks/calendar-day.scss
```diff
@@ -0,0 +1,17 @@
+.calendar__day {
+  position: relative;
+  box-sizing: border-box;
```
VinnieJ-2k20, line 3: You have made `day` and element of `calendar` block. That's the right thing to do.
But in this case, styles should be in the same file.

VinnieJ-2k20 (approved):
Good job 👍 
Remember - all style of the block and its elements should be in the same file.

PR #197, opened 2021-07-14
homeworkCalendar

zikozakazabiyaka:
- [DEMO LINK](https://zikozakazabiyaka.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,40 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div class="calendar calendar--start-day--sun calendar--month-length-28">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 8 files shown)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,7 @@
+.page {
```
sasha-morozov, line 1: You placed both outer and inner properties in this selector. Not quite sure that it's allowed according to BEM. Better check it out.

zikozakazabiyaka: fixed, thanks)

src/index.html
```diff
@@ -6,7 +6,40 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div class="calendar calendar--start-day--sun calendar--month-length-28">
```
VinnieJ-2k20, line 10:
```suggestion
    <div class="calendar calendar--start-day--sun calendar--month-length--28">
```

VinnieJ-2k20 (approved):
Good job 👍 
Check modifier syntax.

PR #342, opened 2021-10-19
completed calendar

i-stanko:
- [DEMO LINK](https://i-stanko.github.io/layout_calendar/)

### DIFF
--- a/src/styles/utils/_reset.scss
+++ b/src/styles/utils/_reset.scss
@@ -1,0 +1,5 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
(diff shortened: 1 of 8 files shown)


JamesSunderlanddd (approved):
looks good to me!

src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,5 @@
+* {
+  margin: 0;
+  padding: 0;
```
aionova, line 7: don't use * selector for margins and paddings

aionova (changes requested):
You need to add special classes to the calendar, that will change number of days in the calendar and first day of the month

aionova (approved):

PR #547, opened 2022-02-02
calendar

yahyafullstack01:
- [DEMO LINK](https://yahyafullstack01.github.io/layout_calendar/)
- [TEST REPORT LINK](https://yahyafullstack01.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,76 @@
...
+      font-family: Arial;
+      font-size: 30px;
+      content: counter(countdown);
+    }
+
+    &:hover {
+      background: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+      transition: ease-in-out 0.5s;
```
varseniuk, line 60: You should add the `transition` property to the element itself (not to the pseudo-class).
Because now when you take out the cursor from the cell, it turns back rapidly (without transition)
If you would add the transition to the `calendar__day` the transition will work both ways

varseniuk (approved):

PR #575, opened 2022-02-22
class renamed

divineretribution1:
- [DEMO LINK](https://divineretribution1.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    justify-content: center;
+    align-items: center;
+    height: 100%;
+  }
+
+  &__cell:hover {
+    cursor: pointer;
+    transform: translateY(-20px);
+    background-color: pink;
+    transition: transform 0.5s;
```
vitaliy-martynyuk: move it from the pseudo-class to the class for transition to work when cursor leaves the cell as well

vitaliy-martynyuk (changes requested):

PR #602, opened 2022-03-04
Solution

1sheva1:
DEMO LINK:
https://1sheva1.github.io/layout_calendar/

TEST REPORT LINK:
https://1sheva1.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/styles/blocks/_calendar.scss
+++ b/src/styles/blocks/_calendar.scss
@@ -1,0 +1,64 @@
+.calendar {
+  padding: 10px;
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  justify-content: flex-start;
+  gap: $cell-distance;
+
+  &__day {
+    text-align: center;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+    line-height: $cell-size;
+    box-sizing: border-box;
+    border: 1px solid #000;
+    background-color: #eee;
+    width: $cell-size;
+    height: $cell-size;
+
+    &:hover {
+      transform: translateY(-20px);
+      transition-duration: 500ms;
```
vitaliy-martynyuk: move this property from `:hover` to the class for it to work in both cases(when element is hovered and when cursor leaves it)

vitaliy-martynyuk (changes requested):
amount of days is working incorrectly
Screen Shot 2022-03-07 at 11 27 50 AM

PR #637, opened 2022-03-28
Initital commit ready for review. Also added github name to links in …

MabuzaM:
- [DEMO LINK](https://MabuzaM.github.io/layout_calendar/)
- [TEST REPORT LINK](https://MabuzaM.github.io/layout_calendar/report/html_report/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
+$units:px;
```
Rashaad-Ebrahim: It's not necessary to create a variable for the unit of measurement that you'll be using. 

It would seem counterproductive. I would rather use px in this case.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
...
+$week-days: (
+  "mon": 1,
+  "tue": 90,
+  "wed": 178,
+  "thu": 267,
+  "fri": 355,
+  "sat": 444,
+  "sun": 532
```
Rashaad-Ebrahim: Please reconsider the values that you've used in this variable.

You're proving a set value, which would mean that if you wanted to adjust the size of each day card at a later stage, you would then need to change each individual value. 

Could I suggest creating values that can then be subsituted in, e.g: `$gap: 5px` then substituting it in, `"mon": $gap` `"tue": $gap + $varForCardWidth`.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-wrap: wrap;
+  padding: 30#{$units} 10#{$units} 10#{$units};
+  width: 721#{$units};
```
Rashaad-Ebrahim: Use properly named variables to make all the calculations more clear.

It's not clear how you came to this value. It would be tricky to change the value at a later stage.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    flex-basis: calc(100% / 8);
```
Rashaad-Ebrahim: Line 13: This code is redundant as you later set a width and a height.

Another note, with scss, it's not necessary to use `calc()` as sass does the calculation for you without it. `100% / 8` could have been used.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-wrap: wrap;
+  padding: 30#{$units} 10#{$units} 10#{$units};
```
Rashaad-Ebrahim: The `padding` is affecting the size of the date-block width.

Rashaad-Ebrahim (changes requested):
Your code is very well organised! Have a look at some of the changes I've suggest and make this task look spectacular!

src/styles/blocks/calendar.scss
MabuzaM: Noted Thank you have made some changes please check again.

src/styles/blocks/calendar.scss
MabuzaM: Okay thanx Rashaad have modified

src/styles/blocks/calendar.scss
MabuzaM: Have made some changes thank you

src/styles/utils/_variables.scss
MabuzaM: Struggling to implement this change. Maybe I don't understand your suggestion.

src/styles/utils/_variables.scss
MabuzaM: Okay noted have changed

src/styles/utils/_variables.scss
Rashaad-Ebrahim: On this note, what I'm suggesting is that instead of hardcoding the values, use a variable instead.

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,6 @@
+.page {
+  margin: 0 auto;
+  padding: 50% 0;
+  display: flex;
+  justify-content: center;
```
Rashaad-Ebrahim, line 3: `justify-content: center;` is causing your page to have the center arrangement for the bottom row.

Perhaps use margins to center the calendar horizontally. You also need to think about how to center the calendar vertically.

Rashaad-Ebrahim (changes requested):

src/styles/blocks/page.scss
MabuzaM: Owkay noted

meforoma: you have to set calendar in the middle of the screen => `justify` + `align` supplied with `height` would work good
```suggestion
  display: flex;
  justify-content: center;
  align-items: center;
  height: 100vh;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+  }
+
+  &__day:hover {
+    background-color: pink;
+    cursor: pointer;
+    transform: translateY(-20px);
+    transition: all 0.5s;
+  }
```
meforoma: you're repeating the code
advise you to nest this style under `&__day` above, then it would be more clear where it belongs
also suggest you to use transition not on the :hover state, but on the class - then both hover and un-hover will be smooth.
```suggestion
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+    padding: $dayPadding;
+    @include square (100px, 100px);
+
+    border: 1px solid #000;
+
+    @for $day from 1 through 31 {
+      &:nth-child(#{$day}):before {
+        content: "#{$day}";
+      }
+    }
```
meforoma, line 22: - nesting :hover 
- bubbling transition from :hove up to class
- change `all` to `transform` better avoid using `all` as it slows down the page load time
```suggestion
    }

    transition: transform 0.5s;

    &:hover {
      background-color: pink;
      cursor: pointer;
      transform: translateY(-20px);
    }
```

meforoma (changes requested):

src/styles/blocks/calendar.scss
MabuzaM: Thank you have fixed

src/styles/blocks/calendar.scss
MabuzaM: Thank you. Have fixed

src/styles/blocks/page.scss
MabuzaM: Thank you for this.

meforoma (changes requested):
@MabuzaM please correct day tiles to fit the grid
Snapshot 2022-04-01 11 37 31

MabuzaM:
> @MabuzaM please correct day tiles to fit the grid Snapshot 2022-04-01 11 37 31

Thanx done

src/styles/utils/_variables.scss
```diff
@@ -1,4 +1,4 @@
-$daysPerRow: 7;
+$daysPerRow: 8;
```
meforoma: what-why?
7 days was perfectly correct
```suggestion
$daysPerRow: 7;
```

src/styles/blocks/calendar.scss
```diff
@@ -2,7 +2,6 @@
   width: $daysPerRow * $dayShift;
 
   display: flex;
-  justify-content: center;
```
meforoma: this should fix your whole setup
```suggestion
  justify-self: center;
```

meforoma (changes requested):

src/styles/utils/_variables.scss
MabuzaM: Ooops forgot to revert. Was testing. Thanx

src/styles/blocks/calendar.scss
MabuzaM: Okay thanx noted. Fixing

meforoma (approved):

PR #644, opened 2022-03-30
calendar

iLebo-stack:
https://iLebo-stack.github.io/layout_calendar/
https://iLebo-stack.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -14,10 +14,9 @@
   <body class="page">
     <div
       class="
-        calendar
-        page__calendar
-        calendar--start-day--sun
-        calendar--month-length--29"
+      calendar
+      calendar--start-day--sun
+      calendar--month-length--29"
     >
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -2,11 +2,54 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta
+      name="viewport"
+      content="
+        width=device-width,
+        initial-scale=1.0"
+    >
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div
+      class="
+        calendar
+        page__calendar
+        calendar--start-day--sun
+        calendar--month-length--29"
+    >
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -2,11 +2,54 @@
...
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div
+      class="
+        calendar
+        page__calendar
+        calendar--start-day--sun
+        calendar--month-length--29"
+    >
```
efdalyalcin: this doesn't open in the demo link, i don't know why but there is also github failure error for this.

src/styles/main.scss
```diff
@@ -1,3 +1,48 @@
...
+@import "mixins";
+@import "page";
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $length * 7 + $gap * 6;
+  margin: auto;
+  padding: 10px;
+  gap: 1px;
```
efdalyalcin: I think you missed this one, you can use the variable

src/styles/page.scss
```diff
@@ -0,0 +1,8 @@
+body {
+  margin: 0;
+}
+
+.page {
+  display: flex;
```
efdalyalcin, line 11: you can add justify content and align item here so that there is no spaces between your blocks

efdalyalcin (changes requested):
Well done with the working calendar... I think you just need to make your code cleaner. Use flexbox instead of absolute positioning and you are good to go.

src/index.html
iLebo-stack: The link works fine in my end, the github failure I think is because the no test on this task.

src/styles/main.scss
iLebo-stack: Yep, I'll be on it

src/styles/page.scss
iLebo-stack: Those spaces are present because of the gap rule

iLebo-stack: I'm actually using flexbox not position absolute

src/styles/mixins.scss
```diff
@@ -0,0 +1,17 @@
...
+  font-family: Arial, sans-serif;
+  font-size: 30px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  position: absolute;
+  left: 0;
+  right: 0;
+  top: 0;
+  bottom: 0;
```
efdalyalcin, line 14: I tried to leave a comment for here but apparently i deleted by mistake sorry for that. 

In this part i couldn't understand the necessity of position absolute which then you use it in the `&__day.` When i remove it there happens to be a larger gap in between pieces, so i tried to recommend justify content and align items for `.page`

efdalyalcin (changes requested):
I am sorry for the latest comment, i think somehow i deleted my comment where i was confused.

src/styles/page.scss
efdalyalcin: Actually i meant to leave a comment in the mixin part. Apparently i couldn't leave it sorry for that.

src/styles/mixins.scss
iLebo-stack: The reason I used absolute positioning on the `::before` is so that the element fits the entire space of the box, I used `position: relative;` on `&__day` so that each `::before` element is relative to its box

src/index.html
```diff
@@ -2,11 +2,53 @@
...
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div
+      class="
+      calendar
+      calendar--start-day--sun
+      calendar--month-length--29"
+    >
```
meforoma: you've over-split stuff - one attribute - values should be in one line - then no error occurs
```suggestion
    <div class="calendar calendar--start-day--sun calendar--month-length--29">
```

meforoma (changes requested):
your code does not match the deploy - please re-deploy the project

meforoma (approved):

PR #674, opened 2022-04-09
Calendar

dasha-chekmariova:
[DEMO LINK](https://dasha-chekmariova.github.io/layout_calendar/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
+$sizeBlock: 102px;
```
valerii-derkach: why do you use 102px size when in the task size is 100px?

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,12 @@
+@import "../utils/variables";
```
valerii-derkach, line 1: when you @import variables you should not do it to each file, it is enough to @import everything to main.scss file
but when you @use, then use it in every file you need

src/styles/blocks/main.scss
```diff
@@ -0,0 +1,47 @@
+@import "./../utils/variables";
+
```
valerii-derkach: the main.scss file is to @import all styles together
you should move .calendar styles to separate file ./styles/blocks/calendar.scss
the file main.scss should be in folder /styles/main.scss
files normalize.scss and reset.scss shold be in folder ./styles/
folder 'utils' is for variables, functions, mixins

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.page {
+
+  &__calendar {
+    display: flex;
+    flex-wrap: wrap;
+    gap: $gapX;
+    width: 7 * $sizeBlock + 7 * $gapX;
+    margin: 0 auto;
```
valerii-derkach, line 10: 10. [STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.

valerii-derkach (changes requested):
fix the structure of styles files and folders

src/styles/blocks/page.scss
dasha-chekmariova: When I use @import in another file, I have an error, so I need to include _variables.scss in each file where needed

src/styles/blocks/main.scss
dasha-chekmariova: i think, name file is not important, i use style.scss to connect all styles

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    transition: 0.5s;
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        width: 100%;
+        height: 100%;
+        text-align: center;
```
denys-cheporniuk: add all these styles outside the loop, leave in the loop styles that depend only on the iteration of the loop

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+        content: "#{$i}";
+      }
+    }
+
+    &:hover {
+      transform: translateY(-20px);
+      background-color: rgb(235, 167, 181);
+      cursor: pointer;
```
denys-cheporniuk: add `cursor: pointer;` on `&__days` not in hover effect

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #690, opened 2022-04-15
calendar

YarinKiri2l:
- [DEMO LINK](https://YarinKiri2l.github.io/layout_calendar/)
- [TEST REPORT LINK](https://YarinKiri2l.github.io/layout_calendar/report/html_report/)


src/styles/calendar.scss
```diff
@@ -0,0 +1,57 @@
+@import url(./_mixin.scss); // why don't work?
+@import url(./variables.scss); // again
+
```
rostyslav-pyliak, line 3: Try to import without url and brackets, just using doublequotes. And import mixins and variables into pages, where you are using them.

rostyslav-pyliak (changes requested):
You forget to implement the modifiers. 

- Implement start-day modifier for the calendar with mon, tue, wed, thu, fri, sat and sun values

1. Use [@each](https://sass-lang.com/documentation/at-rules/control/each) to create all the modifiers

2. The month should start at the correct column (Monday is the 1st, Friday is the 5th)

3. You can just add correct margin-left for the first day

- Add a modifier month-length for the calendar with values 28, 29, 30 and 31 (use @for)

It sets the last day to show (use [nth-child](https://css-tricks.com/how-nth-child-works/))

PR #706, opened 2022-05-08
First commit / calendar created

Kaunin1:
- [DEMO LINK](https://kaunin1.github.io/layout_calendar/)


Kaunin1:
- Build-in linter does not permit line transition of long class names, regretfully;
- Сalendar could be managed by : number of days / week day ;
- Dimensions of calendar could be tansformed by : size of day-cell ;

src/blocks/calendar.scss
```diff
@@ -0,0 +1,61 @@
...
+    flex-wrap: wrap;
+    padding: $mockup-padding;
+    gap: $gap-size;
+    min-width: $mockup-size;
+    max-width: $mockup-size;
+  }
+
+  &__day {
+    @extend %display-set;
+    align-items: center;
```
mgrinko: To understand this code I needed to scroll down, find what is `%display-set` (it was not obvious from the name)
So here I would prefer to see exact rules but not such an extend

src/utils/_variables.scss
```diff
@@ -0,0 +1,16 @@
+$day-size: 100px;
+$day-count: 31;
+$border-size : $day-size / 100;
+$gap-size: $day-size / 100;
```
mgrinko: why does border-size depend on day-size?

src/utils/_variables.scss
```diff
@@ -0,0 +1,16 @@
...
+$mockup-padding: $day-size / 10;
+$calendar-height: 100vh;
```
mgrinko: currently I don't see the usage so I have no idea of limitations of this height
That's why I would prefer not to use variable here at all

src/utils/_variables.scss
```diff
@@ -0,0 +1,16 @@
...
+$mockup-size: ($gap-size * 6) + (7 * $day-size);
```
mgrinko: ($gap-size * 6)
but
(7 * $day-size)

why the order is not the same?
It is confusing

src/utils/_variables.scss
```diff
@@ -0,0 +1,16 @@
...
+$start-shift: $day-size + $gap-size;
+$week-days:
+  "mon" 0,
+  "tue" $start-shift,
+  "wed" $start-shift * 2,
+  "thu" $start-shift * 3,
+  "fri" $start-shift * 4,
+  "sat" $start-shift * 5,
+  "sun" $start-shift * 6;
```
mgrinko: Why not to use just an index instead of repeating `$start-shift`

src/index.html
```diff
@@ -7,6 +7,40 @@
...
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
+        <div class="calendar__day calendar__day--background_grey"></div>
```
mgrinko: Don't add a modifier for the default styles

mgrinko (changes requested):

src/index.html
Kaunin1: Well noted ,  background color is a default setting , thus to be add directly to the element.
Good suggestion.
Have decided to add modifier  to support idea of bem approach.

src/utils/_variables.scss
Kaunin1: There was a dicision to link a border dimension to the cell size  in order to improve appearance of cells of any size. 
Idea was to automate all was it is possible by use of single parameter for all possible proportional modifications - cell size. 
Your approach is clear , agree with one . Direct setting in that case is more obvious and logical.

src/blocks/calendar.scss
Kaunin1: Agree with suggestion. Usability of code  with direct setting of display set is better. 
@Extend was add to the code for purpose of learning, to utilize possibilities of sass as much as possible.

src/utils/_variables.scss
Kaunin1: Good suggestion ! There is a possibility to improve redability of code and reduce probability of human error.

src/utils/_variables.scss
Kaunin1: Good advice ! Agree wtih your approach . 
Formulas was add  to automate all was it possible for experemental purpose . Formulas confuse and  reduce redability of code and harden understanding of limitations in that case.

src/utils/_variables.scss
Kaunin1: Good advice ! Absolutely agree . The way of calculation could be simplified.

src/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+  color: #000;
+}
+
+.calendar {
+  display: flex;
+  justify-content: center;
+  height: $calendar-height;
+  @for $day-number from 1 to $day-count {
+    &--month-length_#{$day-number} {
+      .calendar__day:nth-child(n + #{$day-number + 1}) {
```
mgrinko: This is not a day-number
It is number-of-days!

src/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  @each $day, $index in $week-days {
+    &--start-day_#{$day} {
+      .calendar__day:nth-child(1) {
+        margin-left: $start-shift * $index;
+      }
+    }
+  }
+
+  &__mockup {
```
mgrinko: Why do you call it mockup (макет)?

src/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+    gap: $gap-size;
+    min-width: $mockup-size;
+    max-width: $mockup-size;
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    @include square($day-size);
```
mgrinko, line 42: Do you use it somewhere else?

src/utils/_variables.scss
mgrinko: I am not about a formula but about `100vh` in a variable

mgrinko (approved):

src/blocks/calendar.scss
Kaunin1: just wanted to use something different form 'container'.
Mockup means that something is on basic way preparation or used as base for specific modifications. Agree that such name is confusing and  not applicable here .

src/blocks/calendar.scss
Kaunin1: No , it was used just ones  - for purpose of practice  in sass :))

Kaunin1: regarding $calendar-height: 100vh -  I was excited by possbilities of sass and tried to implemet as many variables, extends , mixins as possible . For sure - that is should be  direct setting of the field.

src/utils/_variables.scss
Kaunin1: regarding $calendar-height: 100vh - I was excited by possibilities of sass and tried to implemet as many variables, extends , mixins as possible . For sure - that has to be a direct setting of the element , nobody goes to change that setting ever ! 8-)

src/blocks/calendar.scss
Kaunin1: Agree ! a bit  longer name, but more clear .

src/blocks/calendar.scss
Kaunin1: just tried to use something different from 'container'.  Mockup of project - means that  it 's readiness on basic level and  ready for further specific modification.  Agree , that  Mockup term is not applicable in our case,  meaning is quite confusing.

PR #955, opened 2022-07-19
calendar solution

ibondrk:
- [DEMO LINK](https://ibondrk.github.io/layout_calendar/)


SergeyPistelev (approved):
Great job! 
Respect for the sparkline)

src/index.html
```diff
@@ -7,6 +7,50 @@
...
+      <div class="sparkline__day">mon</div>
+      <div class="sparkline__day">tue</div>
+      <div class="sparkline__day">wed</div>
+      <div class="sparkline__day">thu</div>
+      <div class="sparkline__day">fri</div>
+      <div class="sparkline__day">sat</div>
+      <div class="sparkline__day">sun</div>
+    </div>
+
+    <div class="calendar calendar--month-length_31">
```
dellachaise: Rename modifiers following BEM naming convention for complex modifiers: `block-name--modifier-name--modifier-value` (checklist)

src/index.html
```diff
@@ -7,6 +7,50 @@
...
+      <div class="calendar__field">
+        <div class="calendar__day calendar__day--start-day_fri"></div>
```
dellachaise: It's a `calendar` block modifier, move it to the calendar classes. And rename following BEM naming convention for complex modifiers.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  position: relative;
+  @extend %flex-view--center;
+  @extend %font;
+  @extend %calendar-width;
+  margin: auto;
```
dellachaise: Don't add margin and position to a block, only for element. Use mix.
 More details in video.
[video, timing 4:15](https://mate.academy/learn/css-advanced/bem#/video/911)
Screenshot 2022-07-22 at 12 30 04

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+    @for $index from 1 through 31 {
+      &:nth-child(#{$index}) {
+        &::before {
+          content: "#{$index}";
+        }
+      }
+    }
+    &:hover {
+      cursor: pointer;
```
dellachaise: move cursor: pointer to &__day.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+      transform: translateY(-20px);
+      transition: 0.5s;
```
dellachaise: add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

Always add transition property. By default, transition property is all, and it's hard for the browser to control all properties.

src/styles/blocks/calendar_sparkline.scss
```diff
@@ -0,0 +1,20 @@
+.sparkline {
```
dellachaise: rename file to a `sparkline`, the same as block name.

src/styles/blocks/calendar_sparkline.scss
```diff
@@ -0,0 +1,20 @@
...
+
+    @for $days from 1 through 7 {
+      &:nth-child(#{$days}) {
+        width: $cell-size;
+        height: 50px;
+        text-transform: uppercase;
+        background-color: #eee;
+        border: 1px solid #000;
+      }
+    }
```
dellachaise: Why do you use cycle for here?

dellachaise (changes requested):
Good start 👍

Please, add fixes according to comments.

src/index.html
ibondrk: I already renamed modifier for first day position, and actually it`s modifier exactly for the first day in month, month could change for any another in this case, so we had to change only first day position. 

In everything else i agree with you fully

dellachaise: I understand that you already implemented start-day modifier this way, but according to the task, `start-day` should be modifier for the `calendar`, not for `calendar__day`.


Screenshot 2022-07-22 at 15 09 09

dellachaise (approved):
Well done 👍 

Read a comment about `start-day` modifier 👆

PR #959, opened 2022-07-19
Develop

VolodymyrChuyko:
- [DEMO LINK](https://VolodymyrChuyko.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day_sun calendar--month-length_29">
+      <div class="calendar__container">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day_sun calendar--month-length_29">
```
Jamal-Amoian, line 10: More correctly is `calendar--start-day--sun` and `calendar--month-length--29`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+      display: none;
+    }
+  }
+
+  &__container {
+    display: flex;
+    flex-wrap: wrap;
+    gap: $gap-size;
+    width: $container-size;
+    min-width: $container-size;
```
Jamal-Amoian: I think, correctly `max-width`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+    padding: 10px;
+
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
```
Jamal-Amoian, line 30: You need to write it to `normalize.scss` into `* {box-sizing: border-box;}`

src/styles/main.scss
```diff
@@ -1,3 +1,9 @@
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
```
Jamal-Amoian, line 3: That's not reset styles, you have to write this in `normalize.scss` file

Jamal-Amoian (changes requested):

src/index.html
VolodymyrChuyko: Thank you for reviewing my code. You're right here if you mean two dashes style, recommended by BEM methodology. But I consciously follow the naming scheme, offered by Mate academy. It looks like that: block__element--modifier_value. I think it is more comfortable, as it provides a better visual difference between modifiers and values. So, if you don't mind, I would like to leave it without changes

src/styles/blocks/calendar.scss
VolodymyrChuyko: Oh, it looks, like redundancy. It doesn't affect anything. I'm just going to delete it. Thanks for bringing this to my attention

src/styles/blocks/calendar.scss
VolodymyrChuyko: This is the only place I use this property in the code. I don't think it is necessary to create a normalize.scss file. So, if you don't mind, I would leave it without changes

src/styles/main.scss
VolodymyrChuyko: I got only a few rows of code for normalization, so I decided to put it in main.scss, not to create a normalize file. As I know, it's a widespread practice. Would you insist on creating normalize file?

Jamal-Amoian: No problem, mentors will judge us

Jamal-Amoian (approved):

danylo-onishchuk (approved):

PR #972, opened 2022-07-20
Create calendar with transformation

max-kravchenko:
- [DEMO LINK](https://max-kravchenko.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,39 @@
...
+  &__cell {
+    @extend %cell-formatting;
+    min-width: $cell-width;
+    min-height: $cell-height;
+    transition: color 0.5s, transform 0.5s;
+
+    @for $i from 1 through 31 {
+      &:nth-child(n + #{$i})::before {
+        content: "#{$i}";
+      }
```
NataliyaAn, line 29: Nicely done)

src/styles/extends.scss
```diff
@@ -0,0 +1,21 @@
...
+  justify-items: center;
+  align-items: center;
+}
+
+%calendar-formatting {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  justify-items: center;
+  margin: 100px auto;
```
NataliyaAn: in my opinion - task 'Display a calendar in the middle of the screen' is matters in the middle of height and in the middle of width. you did margin top and bottom 100px. it's not the same

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translate(0, -20px);
+      transition: 0.5s;
```
NataliyaAn: this could be done through animation. but it wasn't necessary

NataliyaAn (approved):

src/styles/extends.scss
max-kravchenko: seems like I used some cheats :(

src/styles/blocks/calendar.scss
max-kravchenko: thanks!!

danylo-onishchuk (approved):

PR #976, opened 2022-07-21
Created calendar

OlgaOchichenko:
[DEMO LINK](https://OlgaOchichenko.github.io/layout_calendar/)
 [TEST REPORT LINK](https://OlgaOchichenko.github.io/layout_calendar/report/html_report/)


src/styles/blocs/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+  background-color: #eee;
+  height: $day-size;
+  width: $day-size;
+  border: 1px solid #000;
+  transition: color 0.5s, transform 0.5s;
+
+  &:hover {
+    background-color: #ffc0cb;
+    cursor: pointer;
+  }
```
ghost: You forgot to add **transition** to **hover**.
That was said in the task: **The hovered cell has to become pink
(Optional) Move it up by 20px (use transform)
(Optional) both of these properties are to be animated with the duration of half a second.**

ghost (changes requested):
If I were you, I would transfer the variables file into another folder, for example, **utils** or something like that. Because the variables aren't the blocks.  
Also in the checklist was said:  **[STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.**
Your calendar is good! You should just correct these small mistakes.
If you will have any questions or maybe you disagree with something, don't hesitate to write me on slack. I will be happy to help)

ghost (approved):

AlinaMatuschak (changes requested):
try run deploy one more time 
Screenshot 2022-07-21 at 22 57 11

AlinaMatuschak (changes requested):
- class calendar--month-length--29 doesn’t work 
- should display a calendar in the middle of the screen
Screenshot 2022-07-22 at 20 14 29

danylo-onishchuk (approved):

PR #1043, opened 2022-08-09
Develop

Proph7000:
https://proph7000.github.io/layout_calendar/
https://proph7000.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,11 +1,3 @@
-* {
-  box-sizing: border-box;
-}
-
-body {
-  margin: 0;
-}
-
 .calendar {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  &__day:hover {
+    background-color: pink;
+    transform: translateY(-20px);
+    transition: 0.5s;
+  }
+
+  $name-of-the-day:
```
alexandra-protyanova, line 35: it is better to remove map to  separate file

alexandra-protyanova (approved):
Good job!

PR #1060, opened 2022-08-16
Calendar

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/layout_calendar/)

But there is no html report for this task))
- [TEST REPORT LINK](https://MaxManis.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
+      <div class="calendar__box calendar__box--month-length--29 calendar__box--start-day--sun">
+        <div class="calendar__day calendar__day--day--1"></div>
+        <div class="calendar__day calendar__day--day--2"></div>
+        <div class="calendar__day calendar__day--day--3"></div>
+        <div class="calendar__day calendar__day--day--4"></div>
+        <div class="calendar__day calendar__day--day--5"></div>
+        <div class="calendar__day calendar__day--day--6"></div>
+        <div class="calendar__day calendar__day--day--7"></div>
+        <div class="calendar__day calendar__day--day--8"></div>
+        <div class="calendar__day calendar__day--day--9"></div>
+        <div class="calendar__day calendar__day--day--10"></div>
+        <div class="calendar__day calendar__day--day--11"></div>
+        <div class="calendar__day calendar__day--day--12"></div>
+        <div class="calendar__day calendar__day--day--13"></div>
+        <div class="calendar__day calendar__day--day--14"></div>
+        <div class="calendar__day calendar__day--day--15"></div>
+        <div class="calendar__day calendar__day--day--16"></div>
+        <div class="calendar__day calendar__day--day--17"></div>
+        <div class="calendar__day calendar__day--day--18"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
```
Anna-Borodaienko, line 10: I'd recommend to leave this header, but make it invisible

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,74 @@
...
+    //position: relative;
+
+    //Main Alignment Method:
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+
+    height: $day-size;
+    width: $day-size;
```
Anna-Borodaienko: Or you can use mixin here

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,74 @@
...
+.calendar {
+
+  &__box {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 1px;
+
+    padding-top: 40px;
+
+    width: $calendar-width;
```
Anna-Borodaienko: You don't need blank line here, both are one section, block

src/index.html
MaxManis: i guess its useless so i removed it as i always do)

OleksiiBlezniuk: Agree with @MaxManis , I don't see any sense in it

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,71 @@
...
+    //position: relative;
+
+    //Main Alignment Method:
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+
+    height: $day-size;
+    width: $day-size;
```
OleksiiBlezniuk: use mixin

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,71 @@
...
+    border: 1px solid black;
+    background-color: #eee;
```
OleksiiBlezniuk: use scss variables for colors

src/styles/variables.scss
```diff
@@ -0,0 +1,14 @@
...
+$calendar-width: $day-size * 7 + 6;
+$days: (
+  "mon": 0,
+  "tue": 101px,
+  "wed": 202px,
+  "thu": 303px,
+  "fri": 404px,
+  "sat": 505px,
+  "sun": 606px
+);
```
OleksiiBlezniuk: Strange solution, why hardcode exact size for each case, set day number here and then just calculate needed value

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,71 @@
...
+
+    transition: color 0.5s, background-color 0.5s, transform 0.5s;
```
OleksiiBlezniuk: use variable for transition-duration

OleksiiBlezniuk (changes requested):
GJ, left few comments to fix

MaksimusCRS (approved):
GJ

PR #1065, opened 2022-08-18
readme.md was modified, calendar task was added

TheVovchik:
- [DEMO LINK](https://TheVovchik.github.io/layout_calendar/)
- [TEST REPORT LINK](https://TheVovchik.github.io/layout_calendar/report/html_report/)


src/styles/elements/_variables.scss
```diff
@@ -0,0 +1,13 @@
...
+$month-width: 7 * $day-sides + 2 * $block-padding + 6px;
+$day-color: #eee;
+$start-days: (
+  "mon" 0 7,
+  "tue" 101px 6,
+  "wed" 202px 5,
+  "thu" 303px 4,
+  "fri" 404px 3,
+  "sat" 505px 2,
+  "sun" 606px 1
```
Anna-Borodaienko: Strange solution, better set here modifier for day number and save one cell size, than multiply them, I don't think that it's a good idea to hardcode size for every case 🙂
I had the same solution - and mentor didn't like it

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  width: $month-width;
+  padding: 10px;
+
+  @for $days-count from 28 through 31 {
+    &--month-length--#{$days-count} {
+      & :nth-child(1n + #{ $days-count + 1 }) {
```
Anna-Borodaienko:
```suggestion
      & :nth-child(n + #{ $days-count + 1 }) {
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+      // }
+      // & :nth-child(7n + #{$weekend - 1}) {
+      //   background-color: rgba(100%, 0, 0, 0.6);
+      // }
+    }
+  }
+
+  &__day {
+    height: $day-sides;
+    width: $day-sides;
```
Anna-Borodaienko, line 26: Use here mixin

src/index.html
```diff
@@ -1,12 +1,49 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
+      <div class="calendar calendar--month-length--29 calendar--start-day--sun">
+        <div class="calendar__day calendar__day--day-1"></div>
```
Anna-Borodaienko:
```suggestion
        <div class="calendar__day calendar__day--day--1"></div>
```
The value of a modifier is separated from its name by a double hyphen (--).

src/styles/elements/_variables.scss
TheVovchik: reworked with index of the day

src/styles/blocks/calendar.scss
TheVovchik: Not optimal desicion if you have rule only once

src/index.html
TheVovchik: God please save BEM)))

src/styles/blocks/calendar.scss
TheVovchik: not like mistake, its work like unnecessary modifier) get it out)

src/index.html
```diff
@@ -1,12 +1,49 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
+      <div class="calendar calendar--month-length--29 calendar--start-day--sun">
+        <div class="calendar__day calendar__day--day--1"></div>
```
OleksiiBlezniuk: I believe that you could complete this task without this modifiers for every element 🙂

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+        margin-left: $first-day + $first-day * $day-sides;
+      }
+    }
+  }
+
+  &__day {
+    height: $day-sides;
+    width: $day-sides;
+    background-color: $day-color;
+    border: 1px solid black;
```
OleksiiBlezniuk, line 28: use variables for colors

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    transition: all 0.5s;
```
OleksiiBlezniuk, line 29: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

src/styles/elements/_variables.scss
```diff
@@ -0,0 +1,14 @@
+$day-sides: 100px;
+$block-padding: 10px;
+$calendar-gap: 1px;
+$month-width: 7 * $day-sides + 2 * $block-padding + 6 * $calendar-gap;
```
OleksiiBlezniuk, line 4: Better to save days amount in variables, now it looks like magic numbers

OleksiiBlezniuk (approved):

src/index.html
TheVovchik: you god damn right!

src/styles/blocks/calendar.scss
TheVovchik: don't like to use variables for default colors)

src/styles/blocks/calendar.scss
TheVovchik: agry!

src/styles/blocks/calendar.scss
OleksiiBlezniuk: Usually there are few colours on the project (for example 10 or less) and almost all of them are different from default. At some time designers can decide to change them a bit, and variables will make this changes easily, because it's painful to find every particular color usage to change it 🙂  That's why we recommend to use variables even in small projects (to get used to this logic)

src/index.html
OleksiiBlezniuk: image

PR #1075, opened 2022-08-20
add calendar

AndriyZhabynets:
- [DEMO LINK](https://AndriyZhabynets.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -32,7 +32,7 @@
 
     @each $day, $number in $days {
       .calendar--start-day--#{$day} &:first-child {
-        margin-left: $number;
+        margin-left: $number * ($square-size + $calendar-gap);
       }
     }
 
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,45 @@
+.calendar {
+  display: flex;
+  width: getCalendarWidth($square-size, $calendar-padding);
+  flex-wrap: wrap;
+  margin: auto;
+  padding-top: 80px;
+  gap: 1px;
+
+  &__day {
+    display: flex;
+    justify-content: center;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+          content: "#{$i}";
+        }
+      }
+    }
+
+    @each $number in $num-day-in-month {
+      .calendar--month-length--#{$number} &:nth-child(n + #{$number + 1}) {
+        display: none;
+      }
+    }
```
Yeliseiev, line 31: In condition they offer here use @for

Yeliseiev (approved):

src/styles/blocks/calendar.scss
AndriyZhabynets: Ok, but i think it is not necessarily

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,14 @@
...
+$calendar-gap: 1px;
+
+$days:
+  "mon" 0,
+  "tue" 1 * ($square-size + $calendar-gap),
+  "wed" 2 * ($square-size + $calendar-gap),
+  "thu" 3 * ($square-size + $calendar-gap),
+  "fri" 4 * ($square-size + $calendar-gap),
+  "sat" 5 * ($square-size + $calendar-gap),
+  "sun" 6 * ($square-size + $calendar-gap);
```
Anna-Tkachenko: you need to store only week day and the number
than go through this variable with cycle, and set correct margins to the needed selectors (formula should be written only once in the cycle, don't repeat it many times)

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1088, opened 2022-08-22
Calendar was created

Yevheniya-Deryka:
- [DEMO LINK](https://Yevheniya-Deryka.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Yevheniya-Deryka.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -24,6 +24,8 @@
     border: 1px solid black;
     background-color: $day-bcg-color;
 
+    cursor: pointer;
+
     transition: background-color 0.5s, transform 0.5s;
 
     @for $index from 1 through 31 {
@@ -31,8 +33,6 @@
         content: "#{$index}";
 
         text-align: center;
-        font-family: Arial, Helvetica, sans-serif;
-        font-size: 30px;
       }
     }
 
@@ -45,7 +45,6 @@
     &:hover {
       transform: translateY(-20px);
       background-color: $hovered-day-bcg-color;
-      cursor: pointer;
     }
   }
 }
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -5,8 +5,7 @@
   gap: $calendar-gap;
 
   width: $calendar-width;
-  padding: $calendar-paddings;
-  padding-top: 40px;
+  padding: 30px $calendar-paddings $calendar-paddings;
 
   @for $month-end from 28 through 31 {
     &--month-length--#{$month-end} > :nth-child(n + #{$month-end + 1}) {
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: $calendar-gap;
+
+  width: $calendar-width;
+  padding: $calendar-paddings;
+  padding-top: 40px;
+
+  @for $month-end from 28 through 31 {
+    &--month-length--#{$month-end} > :nth-child(n + #{$month-end + 1}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    @include square($day-side-length);
+
+    border: 1px solid black;
+    background-color: $day-bcg-color;
+
+    transition: background-color 0.5s, transform 0.5s;
+
+    @for $index from 1 through 31 {
+      &--date--#{$index}::before {
+        content: "#{$index}";
+
+        text-align: center;
+        font-family: Arial, Helvetica, sans-serif;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: $calendar-gap;
+
+  width: $calendar-width;
+  padding: $calendar-paddings;
+  padding-top: 40px;
```
TheVovchik: you have paddings on in variables = 10px, but you want to make top 40px, so may be type
padding: 40px 10px 10px; here it will be like 40px top, 10px left-right(sides) 10px bottom

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+    transition: background-color 0.5s, transform 0.5s;
+
+    @for $index from 1 through 31 {
+      &--date--#{$index}::before {
+        content: "#{$index}";
+
+        text-align: center;
+        font-family: Arial, Helvetica, sans-serif;
+        font-size: 30px;
```
TheVovchik: move  FF and FS to HTML so they will be inherited here (it's not a fault)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    @each $day, $number in $days-of-the-week {
+      &--start-day--#{$day} {
+        margin-left: (#{$number * ($day-side-length + $calendar-gap)});
+      }
+    }
+
+    &:hover {
+      transform: translateY(-20px);
+      background-color: $hovered-day-bcg-color;
+      cursor: pointer;
```
TheVovchik: cursor is not animated, so put him to an element

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,7 @@
+.page {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  margin: 0;
```
TheVovchik, line 6: add - "height: 100vh;"  to make your calendar on the center of the page

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin square ($size) {
+  width: $day-side-length;
+  height: $day-side-length;
```
TheVovchik: you give in mixin variable $size, but use $day-side-length, problem is when I type like an example @include square(40px) it will not work at all. 
you need use in mixin exactly the same variable you declare = $size. it works like $size = variable that i transfer;
in your case it works only because you have $day-side-length declared

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,19 @@
+$day-bcg-color: #eee;
+$hovered-day-bcg-color: pink;
```
TheVovchik, line 2: both variables used only once

TheVovchik (changes requested):

src/styles/blocks/calendar.scss
Yevheniya-Deryka: can't agree with you) 
I use the variable 'calendar-paddings' several times (to count the width of the container, for example). 
and this one top-padding I use only here. maybe I can make one more variable for padding-top, but it doesn't have sense IMHO

src/styles/blocks/calendar.scss
Yevheniya-Deryka: agree

src/styles/utils/_mixins.scss
Yevheniya-Deryka: yep!) will fix that

src/styles/utils/_variables.scss
Yevheniya-Deryka: I've been training to use variables ))

src/styles/blocks/calendar.scss
TheVovchik: I more tell that here isn't necessary to use any variable in this case. just use one line, and define padding once. now u have 2 lines, and one line change properties of top one. I just tell that if you have some style value only once it's better to type it, than create variable.

Yevheniya-Deryka: ok. but in that case, I'll need to type 10px into the count of calendar width manually too, right?

TheVovchik: no

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,19 @@
...
+$day-side-length: 100px;
+
+$calendar-gap: 1px;
+$calendar-paddings: 10px;
+$calendar-width:
+  $day-side-length * 7
+  + $calendar-gap * 6
+  + $calendar-paddings * 2;
```
MaksimusCRS, line 10: Better to put all numbers to the variables, then you would have clean code

MaksimusCRS (approved):

PR #1089, opened 2022-08-22
calendar created

Yeliseiev:
- [DEMO LINK](https://Yeliseiev.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Yeliseiev.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -2,8 +2,9 @@
 @import "./utils/variables";
 
 .calendar {
-  // Padding просто від балди написаний, щоб не приліпало до екрану.
-  padding: 60px;
+  margin: auto;
+  padding-top: 60px;
+  
   display: flex;
   flex-wrap: wrap;
   gap: 1px;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,10 +1,4 @@
-@import "./utils/mixins";
-@import "./utils/variables";
-
 .calendar {
-  margin: auto;
-  padding-top: 60px;
-  
   display: flex;
   flex-wrap: wrap;
   gap: 1px;
@@ -40,10 +34,10 @@
     }
   }
 
-  @each $day-name, $margin-left in $day-of-week {
+  @each $day-name, $day-start in $day-of-week {
     &--start-day-#{$day-name} {
       .calendar__day:nth-child(1) {
-        margin-left: $margin-left;
+        margin-left: ($day-start * ($day-size + $days-gap));
       }
     }
   }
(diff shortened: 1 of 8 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
+@import "./utils/mixins";
+@import "./utils/variables";
+
+.calendar {
+  // Padding просто від балди написаний, щоб не приліпало до екрану.
+  padding: 60px;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
```
AndriyZhabynets: You must display a calendar in the middle of the screen, it is a task demand

AndriyZhabynets (commented):
Good job, but you should to fix one thing

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$day-size: 100px;
+$row-days: 7;
+$days-gap: 1px;
+$padings: 10px;
+
+$calendar-width: $row-days * $day-size + $padings * 2;
+
```
AndriyZhabynets, line 7: i think you can use function in separate file for calculate this number

AndriyZhabynets (changes requested):

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
+@import "./utils/mixins";
+@import "./utils/variables";
+
+.calendar {
+  margin: auto;
```
Anna-Tkachenko: you can't set margins to BEM block 🥲

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$day-of-week:
+  "mon" 0,
+  "tue" 1 * ($day-size + $days-gap),
+  "wed" 2 * ($day-size + $days-gap),
+  "thu" 3 * ($day-size + $days-gap),
+  "fri" 4 * ($day-size + $days-gap),
+  "sat" 5 * ($day-size + $days-gap),
+  "sun" 6 * ($day-size + $days-gap);
```
Anna-Tkachenko: you can store only week day and number of day
and create cycle that will go through the days and set correct margins to needed selectors
so you shouldn't repeat this formula many times

Anna-Tkachenko (changes requested):

src/styles/utils/_variables.scss
Anna-Tkachenko: this calculation used only one time, so it's not required to create separate function for it

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1095, opened 2022-08-22
Calendar was created

anastasiia-nahaiska:
- [DEMO LINK](https://anastasiia-nahaiska.github.io/layout_calendar/)

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,8 +1,8 @@
-$cell-wiidth: 100px;
+$cell-width: 100px;
 $calendar-internal-gap:1px;
 $calendar-outer-gap: 10px;
 $calendar-width:
-  $cell-wiidth * 7 +
+  $cell-width * 7 +
   $calendar-internal-gap * 6 +
   $calendar-outer-gap * 2;
 $day-of-week:
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,0 +1,15 @@
+$cell-wiidth: 100px;
+$calendar-internal-gap:1px;
+$calendar-outer-gap: 10px;
+$calendar-width:
+  $cell-wiidth * 7 +
+  $calendar-internal-gap * 6 +
+  $calendar-outer-gap * 2;
+$day-of-week:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -9,8 +9,8 @@
     display: flex;
     justify-content: center;
     align-items: center;
-    height: 100px;
-    width: $cell-wiidth;
+    height: $cell-width;
+    width: $cell-width;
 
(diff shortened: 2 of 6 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  flex-wrap: wrap;
+  gap: $calendar-internal-gap;
+  padding: 200px $calendar-outer-gap 0;
+  max-width: $calendar-width;
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100px;
```
a-nahaitsev: you can use valiable that you used for width, because it's square, and give a name, for example, $cell-side

a-nahaitsev (changes requested):
gj! 0i have only one comment

src/styles/calendar.scss
anastasiia-nahaiska: Oh, really! Thanks!

a-nahaitsev (approved):

src/styles/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$cell-width: 100px;
+$calendar-internal-gap:1px;
+$calendar-outer-gap: 10px;
+$calendar-width:
+  $cell-width * 7 +
```
OleksiiBlezniuk, line 5: Better to save amount of days in variable, now it looks like magic number

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: $cell-width;
+    width: $cell-width;
+
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
+
+    background-color: #eee;
```
OleksiiBlezniuk, line 18: Use variables for colors

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    border: 1px solid #000;
+
+    transition: ease 0.5s;
```
OleksiiBlezniuk, line 21: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+      &:nth-child(#{$date})::before {
+        content: "#{$date}";
+      }
+    }
+  }
+
+  @each $day, $number in $day-of-week {
+    &--start-day--#{$day} {
+      & :first-child {
+        margin-left: $number * ($cell-width + $calendar-internal-gap);
```
OleksiiBlezniuk, line 39: Would be nice to move it to function

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: $cell-width;
+    width: $cell-width;
+
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
```
OleksiiBlezniuk, line 16: Set font styles to `html` tag

OleksiiBlezniuk (approved):

PR #1098, opened 2022-08-22
Develop calendar

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,40 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap;
+
+  max-width: $calendar-width;
+  padding: 50px $horizontal-padding;
+
+  &__day {
+    @include day($cell-size);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+
+    background-color: #eee;
+    cursor: pointer;
+
+    @for $day from 1 through 31 {
+      &:nth-child(#{$day})::before {
+        content: '#{$day}'
+      }
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+  padding: 50px $hPadding;
+
+  &__day {
+    @include day($cellSize);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+
+    background-color: #eee;
```
xiimiik: As for me, it's better to put all the colors in a variable

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+
+  max-width: $calendarWidth;
+  padding: 50px $hPadding;
+
+  &__day {
+    @include day($cellSize);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
```
xiimiik: This property does nothing

xiimiik (commented):
Greate code, GJ!

src/styles/blocks/calendar.scss
imaksweb: > As for me, it's better to put all the colors in a variable

Yeah, for sure, but we use each color only once so I see no need of doing that in this task

src/styles/blocks/calendar.scss
imaksweb: > This property does nothing

Yep) completely agree with you

OleksiiBlezniuk (approved):
Good job! 🔥

PR #1101, opened 2022-08-23
deployed layout calendar

dmytristepanov:
- [DEMO LINK](https://dmytristepanov.github.io/layout_calendar/)
- [TEST REPORT LINK](https://dmytristepanov.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/_calendar.scss
+++ b/src/styles/blocks/_calendar.scss
@@ -1,0 +1,54 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: getMaxWidth(
+    $day-box-size, $day-box-gap, $day-box-paddings
+  );
+  max-width: getMaxWidth(
+    $day-box-size, $day-box-gap, $day-box-paddings
+  );
+  gap: $day-box-gap;
+  padding: $day-box-paddings;
+  margin: 0 auto;
+
+  @each $day-name, $day-number in $days-of-week {
+    &--start-day--#{$day-name} :first-child {
+      margin-left: getFirstDayPosition(
+        $day-box-size,
+        $day-box-gap,
+        $day-number
+      );
+    }
+  }
+
+  @for $days-value from 28 through 31 {
+    &--month-length--#{$days-value} :nth-child(n + #{$days-value + 1}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    @include day-square($day-box-size);
+
+    background-color: $day-box-color;
+    border: 1px solid black;
+    transition: background-color, transform 0.5s;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: getMaxWidth(
+    $day-box-size, $day-box-gap, $day-box-paddings
+  );
+  max-width: getMaxWidth(
+    $day-box-size, $day-box-gap, $day-box-paddings
+  );
```
glu3s: I am not sure why you assign both `min-width` and `max-width` to the same value. I think setting only the `width` would suffice

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    @include day-square($day-box-size);
+
+    background-color: $day-box-color;
+    border: 1px solid black;
+    transition: background-color, transform 0.5s;
```
glu3s: You need to set the time for each property. Now your background color changes instantly

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    &:hover {
+      cursor: pointer;
+      background-color: $day-box-hover-color;
+      transform: translateY(-20px);
+    }
+
+    @for $day-value from 1 through 31 {
+      &:nth-child(#{$day-value}) {
+        &::before {
```
glu3s: There is no need for extra nesting, you can simply combine these selectors

src/styles/utilities/_functions.scss
```diff
@@ -0,0 +1,7 @@
+@function getFirstDayPosition($box-size, $box-gap, $day-number) {
+  @return ($box-size + $box-gap) * $day-number;
+}
+
+@function getMaxWidth($box-size, $box-gap, $box-paddings) {
+  @return ($box-size * 7) + ($box-gap * 6) + ($box-paddings * 2);
```
glu3s, line 6: Not necessarily a mistake, but if you do not set `box-sizing` to `border-box`, but instead keep it a default
`content-box`, you don't need to account for `$box-paddings`

glu3s (changes requested):
Left some mistakes in the comments

src/styles/utilities/_functions.scss
glu3s: But in that case you would also need to set `border-box` for the day boxes

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1108, opened 2022-08-23
Layout was done

OleksandrGlulhovskiy:
- [DEMO LINK](https://OleksandrGlulhovskiy.github.io/layout_calendar/)
- [TEST REPORT LINK](https://OleksandrGlulhovskiy.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -28,13 +28,12 @@
   @each $day, $number in $days-count {
     &--start-day--#{$day} :first-child {
         margin-left: $number * ($day-size + $gap-width);
-      }
     }
   }
 
   @for $index from 28 through 31 {
     &--month-length--#{$index} :nth-child(n + #{$index + 1}) {
-        display: none;
+      display: none;
     }
   }
 }
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,12 +1,16 @@
 .calendar {
   display: flex;
   flex-wrap: wrap;
-  margin: 0 auto;
-  max-width: $calendar-width;
+  margin: auto;
+  width: $calendar-width;
   gap: $gap-width;
   padding: 40px $calendar-side-paddings 0;
 
   &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
     @include square($day-size);
     border: 1px solid #000;
     background-color: #eee;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -7,23 +7,15 @@
   padding: 40px $calendar-side-paddings 0;
 
   &__day {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    position: relative;
     @include square($day-size);
     border: 1px solid #000;
     background-color: #eee;
-    transition: ease 0.5s;
+    transition: background-color, transform 0.5s;
     cursor: pointer;
 
     &:hover {
       background-color: #ffc0cb;
       transform: translateY(-20px);
-    }
-
-    &::before {
-      position: absolute;
     }
 
     @for $index from 1 through 31 {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0 auto;
```
glu3s: You shouldn't set margins to blocks, do it in a parent block instead

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  max-width: $calendar-width;
```
glu3s: Page behaves strange in the narrow viewport, try it in developer tools. Probably would be better to set `width` instead of `max-width`
```suggestion
  width: $calendar-width;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  gap: $gap-width;
+  padding: 40px $calendar-side-paddings 0;
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
```
glu3s, line 13: It works perfectly fine with the default `static` value. There is no need rewrite it

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    @include square($day-size);
+    border: 1px solid #000;
+    background-color: #eee;
+    transition: ease 0.5s;
```
glu3s: `ease` is a default value for `transition-timing-function`, so you can avoid it, but you should specify the transition properties. In this case that would be `background-color` and `transform`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    cursor: pointer;
+
+    &:hover {
+      background-color: #ffc0cb;
+      transform: translateY(-20px);
+    }
+
+    &::before {
+      position: absolute;
+    }
```
glu3s: The same) It is works alright with the default `static` value

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    @for $index from 1 through 31 {
+      &:nth-child(#{$index})::before {
+        content: "#{$index}";
+      }
+    }
+  }
+
+  @each $day, $number in $days-count {
+    &--start-day--#{$day} {
+      & :first-child {
```
glu3s: You can combine these selectors
```suggestion
    &--start-day--#{$day} :first-child {
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+        margin-left: $number * $day-size + $gap-width * $number;
```
glu3s:
```suggestion
        margin-left: $number * ($day-size + $gap-width);
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  }
+
+  @for $index from 28 through 31 {
+    &--month-length--#{$index} {
+      & :nth-child(n + #{$index + 1}) {
```
glu3s: Same) Can be combined
```suggestion
    &--month-length--#{$index} :nth-child(n + #{$index + 1}) {
```

glu3s (commented):
Left some comments for you

src/styles/blocks/calendar.scss
OleksandrGlulhovskiy: It's the same situation with width

glu3s: Are you sure? Because, when I tried it in the dev tools it's behavior was different

src/styles/blocks/calendar.scss
```diff
@@ -7,25 +7,17 @@
...
   &__day {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    position: relative;
     @include square($day-size);
     border: 1px solid #000;
     background-color: #eee;
-    transition: ease 0.5s;
+    transition: background-color, transform 0.5s;
```
glu3s, line 17: This way you set transition time only for transform property, the background color changes immediately
```suggestion
    transition: background-color 0.5s, transform 0.5s;
```

src/styles/blocks/calendar.scss
OleksandrGlulhovskiy: did't agree, numbers' position on calendars cards break

src/styles/blocks/calendar.scss
OleksandrGlulhovskiy: i made it in calendar block for center it

glu3s (approved):
Nice work!

MaksimusCRS (approved):
GJ

PR #1139, opened 2022-08-25
Develop

NikolayGit1:
- [DEMO LINK](https://nikolaygit1.github.io/layout_calendar/)
- [TEST REPORT LINK](https://nikolaygit1.github.io/layout_calendar/report/html_report/)


src/styles/block/calendar.scss
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  cursor: pointer;
+  padding: 49px;
+  border: 1px solid;
+  background-color: #eee;
+  transition: transform 0.5s;
+
+  @for $index from 29 through 31 {
```
alexandra-protyanova: this loop doesn't work, need to fix)
U are using class for calendar, but don't do it in loop

alexandra-protyanova (changes requested):

PR #1145, opened 2022-08-26
calendar done with sass

AniaHna:
- [DEMO LINK](https://AniaHna.github.io/layout_calendar/)
- [TEST REPORT LINK](https://AniaHna.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sat calendar--month-length-30">
+      <div class="day day--1"></div>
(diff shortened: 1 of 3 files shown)


Anna-Tkachenko (changes requested):
calendar length doesn't change when I change the modifier, fix it
image

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sat calendar--month-length-30">
+      <div class="day day--1"></div>
```
Anna-Tkachenko, line 11: will be better to use nth-child instead of creating separate class for each day

Anna-Tkachenko (approved):

PR #1159, opened 2022-08-28
Develop calendar

constantweb:
[DEMO LINK](https://constantweb.github.io/layout_calendar/)

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,9 +1,8 @@
 $block-size: 100px;
-$block-padding: 10px;
 $block-gap: 1px;
 $block-bg: #eee;
-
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$block-size: 100px;
+$block-padding: 10px;
```
qoa11a: If you have block-size that equals to width and height of the block than it's logical to suppose that block-padding is the padding of this block. But you use it as padding of calendar. You should probably name it calendar-padding than.

qoa11a (changes requested):

qoa11a (approved):

PR #1183, opened 2022-08-31
src

ShoxruxAxmedov:
- [DEMO LINK](https://ShoxruxAxmedov.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,7 @@
-body {
-  margin: 0;
-}
+@import "../styles/block/calendar.scss";
+@import "../styles/block/days.scss";
+@import "../styles/block/month.scss";
+@import "../styles/block/reset.scss";
+
+@import "../styles/utilis/mixin.scss";
+@import "../styles/utilis/variables.scss";
```
hajnaloltyan, line 7: these should be the top imports, specially variables, and they are missing '_' 
and mixin is an empty file

src/styles/block/reset.scss
```diff
@@ -0,0 +1,4 @@
+body {
+  margin: 0;
+  padding: 0;
+}
```
hajnaloltyan, line 4: reset also goes into utilis and needs an '_'

src/styles/block/month.scss
```diff
@@ -0,0 +1,12 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 10px;
+  align-items: center;
+  justify-content: center;
+
+  &--start--day_sun>:first-child {
+    margin-left: 606px;
+  }
```
hajnaloltyan, line 11: You need to cretae an @each loop for the start-day modifiers and use variables for calculating the margin

src/styles/block/days.scss
```diff
@@ -0,0 +1,21 @@
...
+  .days:nth-child(#{$index})::before {
+    content: "#{$index}";
+    font-size: 30px;
+  }
+}
+
+@for $index from 30 through 31 {
+    .days:nth-child(#{$index}) {
+      visibility: hidden;
+    }
```
hajnaloltyan, line 20: it's not just for 30 and 31, but also for 28 and 29
and you're creating the modifiers too with this loop

src/styles/block/calendar.scss
```diff
@@ -0,0 +1,7 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: center;
+  width: 718px;
```
hajnaloltyan, line 6: wrong width and use variables to calculate it

src/index.html
```diff
@@ -7,6 +7,41 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <main class="calendar">
+      <div class="month month--start--day_sun month--month--length-29">
+        <div class="days"></div>
```
hajnaloltyan, line 12: Use BEM, and add modifiers to calendar, like it is mentioned in the task's description

src/styles/block/days.scss
```diff
@@ -0,0 +1,21 @@
...
+  background-color: #eee;
+  box-sizing: border-box;
+}
+
+
+@for $index from 1 through 31 {
+  .days:nth-child(#{$index})::before {
+    content: "#{$index}";
+    font-size: 30px;
+  }
```
hajnaloltyan, line 14: font-family is missing

hajnaloltyan (changes requested):
Hi, I don't see the file where you have added the optional transition/animation on hover. With developer tools I can see it, but not here and also those have no timing added nor they are smooth, and they move to the left on hover and not up. Your calendar isn't centered on the page or the numbers inside each square. And you haven't created with loops the required modifiers. Check again the task's description and the checklist,

PR #1189, opened 2022-09-01
complete the task

barsuchellos:
- [DEMO LINK](https://barsuchellos.github.io/layout_calendar/)


Anna-Tkachenko (changes requested):
calendar length doesn't change when I change the month length in class name
image

naziamoff (approved):

PR #1202, opened 2022-09-09
calendar-task-done

YaTut1901:
- [DEMO LINK](https://YaTut1901.github.io/layout_calendar/)
- [TEST REPORT LINK](https://YaTut1901.github.io/layout_calendar/report/html_report/)


marina-tilniak (changes requested):
Make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

PR #1205, opened 2022-09-10
solution2

matvii1:


src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+
 body {
   margin: 0;
+  height: 100vh;
+  display: grid;
+  place-content: center;
+}
+
+@function shift($n) {
+  @return $n * 100 + px;
```
SanyaBratashchuk: move 100 to the variable

SanyaBratashchuk: and move this function to the utils folder

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+}
+
+$month-length: 31;
```
SanyaBratashchuk: create separate file for vars

SanyaBratashchuk (changes requested):

ArturErmolenko (approved):

PR #1207, opened 2022-09-12
commit

Serrggii:
https://Serrggii.github.io/layout_calendar
https://Serrggii.github.io/layout_calendar/report


marina-tilniak (changes requested):
Make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

marina-tilniak (approved):
Good!

PR #1277, opened 2022-10-05
Develop

AndreyChernysh:
- [DEMO LINK](https://AndreyChernysh.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
-body {
-  margin: 0;
+@import url("./zeroing");
```
sofiia-dovhan: Use SCSS import 
without url();

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+
+$itemSize: 100px;
+$gap: 1px;
+$daysInWeek: 7;
+$calendarPadding: 20px;
+$totalSize: $itemSize + $gap;
```
sofiia-dovhan: You need separate SCSS files for variables and in future for blocks as well))

sofiia-dovhan (commented):
Good Job! :fire:

src/styles/main.scss
mkyloa: agree, let's fix this

src/styles/main.scss
mkyloa: agree

mkyloa (changes requested):
let's make fixes that Sofiia proposed

denys-danyliuk (approved):

PR #1310, opened 2022-10-08
compl-calendar-rev0

serhii-shchuchev:
- [DEMO LINK](https://serhii-shchuchev.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -40,7 +40,7 @@
 
   @for $month-days from 28 through 31 {
     &--month-length-#{$month-days} {
-      .calendar__day:nth-child(n + #{$month-days}) {
+      .calendar__day:nth-child(n + #{$month-days + 1}) {
         visibility: hidden;
       }
     }
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -10,8 +10,7 @@
     display: flex;
     justify-content: center;
     align-items: center;
-    width: $size-cell;
-    height: $size-cell;
+    @include size-cell-day;
     background-color: #eee;
     border: 1px solid black;
     transition: 0.4s;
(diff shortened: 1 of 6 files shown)


qoa11a (changes requested):
image
image
I have one less day in calendar than I have in class. Also because of this I can have 27 days.

serhii-shchuchev:
> image image I have one less day in calendar than I have in class. Also because of this I can have 27 days.

thanks, i'm stuck)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,12 @@
+$gap-cell: 1px;
+$size-cell: 100px;
+$padding-calendar: 10px;
+$widht-calendar: 7 * $size-cell + 6 * $gap-cell + 2 * $padding-calendar;
```
Anastasiia-Svintsova: what do 7 and 6 means? let's create a variable for it. For example $days-in-week
```suggestion
$widht-calendar: $days-in-week * $size-cell + ($days-in-week - 1) * $gap-cell + 2 * $padding-calendar;
```

src/styles/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+  flex-wrap: wrap;
+  padding: $padding-calendar;
+  margin: 75px auto;
+  
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: $size-cell;
+    height: $size-cell;
```
Anastasiia-Svintsova: it would be better to create a mixin for it

Anastasiia-Svintsova (changes requested):
Good! Just fix some comments

src/styles/utils/_variables.scss
serhii-shchuchev: There are values that are not changeable, such as:  number of days in a week,  "number Pi", etc. I don't see the point of setting them as variables. But if the customer or mentor wants it that way, then with pleasure)

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1365, opened 2022-11-02
first version

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/layout_calendar/

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -33,11 +33,12 @@
     height: $cell-size;
     background-color: #eee;
     border: 1px solid black;
-    transition-property: bottom;
+    cursor: pointer;
+    transition-duration: 0.5s, 0.5s;
+    transition-property: bottom, background-color;
   }
 
   &__day:hover {
-    cursor: pointer;
     background-color: pink;
     position: relative;
     bottom: 20px;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -6,23 +6,26 @@
   gap: 1px;
   padding: 10px;
   margin: 0 auto;
-  :nth-child(1) {
-    margin-left: 618px;
-  }
-
-  @for $length from 28 to 32 {
-    &--month-length--#{$length} :nth-child(#{$length}) {
-      content: "#{$length}";
+  
+  @each $day-week, $position in $days-week {
+    &--start-day--#{$day-week} :nth-child(1) {
+      margin-left: $position;
     }
   }
 
-  @for $index from 1 to 32 {
+  @for $index from 1 through 31 {
     & :nth-child(#{$index})::before {
       content: "#{$index}";
       display: flex;
       font-size: 30px;
       line-height: 100px;
       justify-content: center;
+    }
+  }
+  
+  @for $length from $length-of-month to 31 {
+    &--month-length--#{$length-of-month} :nth-last-child(31 - #{$length}) {
+      display: none;
     }
   }
 
@@ -32,12 +35,6 @@
     background-color: #eee;
     border: 1px solid black;
     transition-property: bottom;
-    
-    @each $day-week in $days-week {
-      &--start-day--#{$day-week} {
-        content: "#{$day-week}";
-      }
-    }
   }
 
   &__day:hover {
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -28,8 +28,6 @@
       display: none;
     }
   }
-//   &--month-length--#{$length-of-month}
-// &__day:nth-last-child(-n + #{31 - $length})
 
   &__day {
     width: 100px;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,15 +1,14 @@
 .calendar {
-  font-family: Arial, Helvetica, sans-serif;
-  width: 720px;
+  max-width: $max-wight;
   display: flex;
   flex-wrap: wrap;
-  gap: 1px;
+  gap: $gap;
   padding: 10px;
   margin: 0 auto;
   
   @each $day-week, $position in $days-week {
     &--start-day--#{$day-week} :nth-child(1) {
-      margin-left: $position;
+      margin-left: $position * ($cell-size + 3);
     }
   }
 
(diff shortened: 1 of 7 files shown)


Anastasiia-Svintsova (changes requested):
1. Can't see any changes on hover. The checklist: 
image
Animation
2. These tasks haven't been completed. I can't change the length of the month and the day of the start of the week
image

Vladimir-Zadorozhnyi:
I agree with the first note. I fixed it. But I don't understand the second note, please explain.

loralevitska (changes requested):
hi! you report page don't work 
image

2 part
you need add this classes and we need the ability to change modifiers in devtools and the calendar should change
image

E.g. 
image


If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

Vladimir-Zadorozhnyi:
I have problem with second the cycle( for, month-lengt). Why don't it work? I asked in fe_chat but without result.

изображение

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+      line-height: 100px;
+      justify-content: center;
+    }
+  }
+  
+  @for $length from $length-of-month to 31 {
+    &--month-length--#{$length-of-month} :nth-last-child(31 - #{$length}) {
+      display: none;
+    }
+  }
```
Anastasiia-Svintsova: try smth like that

```suggestion
  @for $length from 28 to 31 {
    &--month-length--#{$length} :nth-child(n + #{$length + 1}) {
      display: none;
    }
  }

```

Anastasiia-Svintsova (changes requested):
1. Add transition on hover
Animation

src/styles/blocks/calendar.scss
Vladimir-Zadorozhnyi: I understood how it work. When I use :nth-chilld, in brackets should be only result.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  font-family: Arial, Helvetica, sans-serif;
```
liliia-livandovskaya: font should be in the `html` selector tag and inherited where needed

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+  width: 720px;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 10px;
+  margin: 0 auto;
+  
+  @each $day-week, $position in $days-week {
+    &--start-day--#{$day-week} :nth-child(1) {
+      margin-left: $position;
```
liliia-livandovskaya:
```suggestion
      margin-left: $number * ($cell-size + $gap);

```

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,4 @@
+$days-week:
+  "mon" 0, "tue" 103px, "wed" 206px, "thu" 309px, "fri" 412px,
+  "sat" 515px, "sun" 618px;
+$length-of-month: 29;
```
liliia-livandovskaya: There should be something like that. It's not the best design to hardcode px for each $days-week:
Screenshot 2022-11-09 at 11 21 02

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  font-family: Arial, Helvetica, sans-serif;
+  width: 720px;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 10px;
+  margin: 0 auto;
```
liliia-livandovskaya:
```suggestion
  display: flex;
  flex-wrap: wrap;
  padding: 10px;
  margin: 0 auto;
  gap: $gap;
  max-width: $max-width;
  adding: $padding * 2 $padding;
```

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments and fix your code according to it

Vladimir-Zadorozhnyi:
> Great job! Pay attention to the comments and fix your code according to it

Thank, so much! I fixed code.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,10 @@
+$cell-size: 100px;
+$gap: 1px;
+$days-in-a-week: 7;
+$padding: 10px;
+$max-wight: $cell-size * $days-in-a-week + 2 
+  *$padding + $gap * ($days-in-a-week - 6);
+$days-week:
+  "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4,
+  "sat" 5, "sun" 6;
```
loralevitska: this is not wrong, but this is how it is customary to write values or function names in imports, if they do not fit on one line, then move each value to a new line. This will help in editing or copying in the future

```suggestion
$days-week:
  "mon" 0,
  "tue" 1,
  "wed" 2,
  "thu" 3,
  "fri" 4,
  "sat" 5,
  "sun" 6;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+      display: none;
+    }
+  }
+
+  &__day {
+    width: $cell-size;
+    height: $cell-size;
+    background-color: #eee;
+    border: 1px solid black;
+    transition-property: bottom;
```
loralevitska: add `0.5s` for smoothness animation
image

loralevitska (changes requested):
nice calendar, it remains to fix the code a little

1. place the calendar in the center of the page
image

src/styles/blocks/calendar.scss
Vladimir-Zadorozhnyi: It is optional from the next topic)))

Vladimir-Zadorozhnyi:
> nice calendar, it remains to fix the code a little
> 
>     1. place the calendar in the center of the page
>        image

I didn't catch, calendar in the center!
изображение

src/styles/utils/variables.scss
Vladimir-Zadorozhnyi: Ok.

liliia-livandovskaya (approved):
Great job! You can just add margin-top so the calendar would look better on hover Screenshot 2022-11-11 at 12 16 39

PR #1373, opened 2022-11-05
calendar was created

superpooperxxx:
- [DEMO LINK](https://superpooperxxx.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/_calendar.scss
+++ b/src/styles/blocks/_calendar.scss
@@ -1,0 +1,50 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  align-content: center;
+  gap: $cellGap;
+  padding: 2 * $containerPaddings $containerPaddings; // cl: 11.
+
+  min-width: getWidth($cellSize, $cellGap, $containerPaddings);
+  max-width: getWidth($cellSize, $cellGap, $containerPaddings);
+    
+  // Position of the first day based on --start-day modifier.
+  @each $day, $dayNum in $days {
+    &--start-day_#{$day} &__day:first-child {
+      margin-left: firstDayPos($cellSize, $cellGap, $dayNum);
+    }
+  }
+
+  // Amount of days to display based on --month-length modifier.
+  // Could use :nth-last-child(-n + #{31 - $days})
+  @for $days from 28 through 31 {
+    &--month-length_#{$days} &__day:nth-child(n + #{1 + $days}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    @include square($cellSize);
+    text-align: center;
+
+    border: 1px solid #000;
+    background: $color-main;
+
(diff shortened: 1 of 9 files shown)


src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+      cursor: pointer;
+    }
+
+    // Add a pseudo-element with number to each day.
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+        font-size: 30px;
+        line-height: $cellSize;
+      }
```
Oleksandr-Filo: it is better to use more understandable names for variables (for example number, dayNumber etc)

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,50 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  align-content: center;
+  gap: $cellGap;
+  padding: 2 * $containerPaddings $containerPaddings; // cl: 11.
+
+  min-width: getWidth($cellSize, $cellGap, $containerPaddings);
+  max-width: getWidth($cellSize, $cellGap, $containerPaddings);
```
Oleksandr-Filo, line 9: no need to specify min width and max width here, they are equal

Oleksandr-Filo (approved):
looks good, just a few tips

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+
+    border: 1px solid #000;
+    background: $color-main;
+
+    transition: background-color 0.5s, transform 0.5s;
+
+    &:hover {
+      background-color: $color-accent;
+      transform: translateY(-20px);
+      cursor: pointer;
```
solaryasha: Let's set the cursor not in the pseudoclass

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // Add a pseudo-element with number to each day.
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+        font-size: 30px;
+        line-height: $cellSize;
```
solaryasha: Do we really need to set these rules in the pseudoelement?
I guess, you should set content only

solaryasha (changes requested):

src/styles/blocks/_calendar.scss
superpooperxxx: done, good point tnx

src/styles/blocks/_calendar.scss
superpooperxxx: done

solaryasha (approved):
Good job !

PR #1374, opened 2022-11-05
layout_calendar task finished

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,7 +9,7 @@
 
   <body>
     <main class="main">
-      <section class="calendar calendar--start-date--sun calendar--month-length--31">
+      <div class="calendar calendar--start-date--sun calendar--month-length--31">
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
@@ -41,7 +41,7 @@
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
-      </section>
+      </div>
     </main>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,42 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <main class="main">
+      <section class="calendar calendar--start-date--sun calendar--month-length--31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -6,7 +6,42 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <main class="main">
+      <section class="calendar calendar--start-date--sun calendar--month-length--31">
```
SalehAkaJim:
```suggestion
      <div class="calendar calendar--start-date--sun calendar--month-length--31">
```

I suggest to use <div> here instead of <section>

> The section tag is used when requirements of two headers or footers or any other section of documents needed. Section tag grouped the generic block of related contents. The main advantage of the section tag is, it is a semantic element, which describes its meaning to both browser and developer.

You can read more about it in [Here](https://www.geeksforgeeks.org/html-section-tag/#:~:text=The%20section%20tag%20is%20used,to%20both%20browser%20and%20developer.)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  &__day {
+    @include flex-positioning(center, center);
+    box-sizing: border-box;
+    width: $day-box-size;
+    height: $day-box-size;
+    border: 1px solid #000;
+    transition: transform 0.5s;
+
```
SalehAkaJim:
```suggestion
    transition: transform 0.5s;
    background-color: #fff;

```
It's better to declare a default value and then change it on :hover...

SalehAkaJim (changes requested):

src/index.html
nurdanokcu: Thank you Saleh. Changes are implemented.

SalehAkaJim (approved):

src/index.html
SalehAkaJim: Happy to help 😅

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,39 @@
...
+  @each $day, $num in $days {
+    &--start-date--#{$day} :first-child {
+      margin-left: ($num - 1) * ($day-box-size + $day-box-gap);
+    }
+  }
+
+  &__day {
+    @include flex-positioning(center, center);
+    box-sizing: border-box;
```
mgrinko, line 21: It would be great to add `cursor: pointer` if we can interact with a day

mgrinko (approved):

PR #1379, opened 2022-11-05
calendar was created

OlggaJul:
- [DEMO LINK](https://olggajul.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,8 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body class="page">
-    <div class="calendar calendar--month-length--31">
-      <div class="calendar__day calendar__day--first-day--mon"></div>
+    <div class="calendar calendar--month-length--31 calendar--first-day--mon">
+      <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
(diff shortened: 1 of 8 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+      display: none;
+    }
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: $dayBlock-size;
+    height: $dayBlock-size;
```
fallowlone, line 25:
```suggestion
    align-items: center;
    
    width: $dayBlock-size;
    height: $dayBlock-size;
```

fallowlone (approved):
I checked your work. It seems good.

src/styles/blocks/calendar.scss
OlggaJul: > 



> I Checked your work. It seems good. But I didn't figure out why you used a variable "weekDays" with the function getMergine.

When I specify a day (eg tue) in html-file, the function calculates the margin knowing that it is the 2nd day of the week. It can be done in another way, but this is what came to my mind.

fallowlone: > 

This is extremely cool.

src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <div class="calendar calendar--month-length--31">
+      <div class="calendar__day calendar__day--first-day--mon"></div>
```
ivan-kadykalo: Would be better to specify a day in the parent tag, as you do it for month=length

ivan-kadykalo (approved):
Good job🚀

PR #1380, opened 2022-11-06
creat calndar

JasterHim:
https://JasterHim.github.io/layout_calendar/

https://JasterHim.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/styles/block/calendar.scss
+++ b/src/styles/block/calendar.scss
@@ -6,28 +6,25 @@
   padding: 0 $calendar__padding;
   gap: $calendar__gap;
 
-  @for $day from 1 to $calendar__month-length {
+  &__day {
+    @include square($calendar__size);
+    box-sizing: border-box;
+    background-color: #eee;
+    border: 1px solid black;
+    text-align: center;
+    line-height: 100px;
+    cursor: pointer;
+    transition: transform 1s;
 
-    &__day {
-      @include square($calendar__size);
-      box-sizing: border-box;
-      background-color: #eee;
-      border: 1px solid black;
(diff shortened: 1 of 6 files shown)


src/styles/block/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+
+    &__day {
+      @include square($calendar__size);
+      box-sizing: border-box;
+      background-color: #eee;
+      border: 1px solid black;
+      text-align: center;
+      line-height: 100px;
+      cursor: pointer;
+      transition: 1s;
```
TarasHoliuk:
```suggestion
      transition: transform 1s;
```
better to specify a property in transition

src/styles/block/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  width: $calendar__width;
+  display: flex;
+  flex-wrap: wrap;
+  margin: 10% auto;
+  padding: 0 $calendar__padding;
+  gap: $calendar__gap;
+
+  @for $day from 1 to $calendar__month-length {
```
TarasHoliuk: this loop is redundant

TarasHoliuk (changes requested):
GJ! But a few fixes are needed

src/styles/block/calendar.scss
vlad-demchuk: fix this comment

vlad-demchuk (changes requested):
check the previous comment

tetiana-ivanchenko (approved):

PR #1390, opened 2022-11-08
initial commit, calendar was added

Anastasia4sik:
- [DEMO LINK](https://Anastasia4sik.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,7 +8,7 @@
   </head>
 
   <body class="page">
-    <main class="calendar calendar--start-day--fri calendar--month-length--30">
+    <main class="calendar calendar--start-day--mon calendar--month-length--31">
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <main class="calendar">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -6,7 +6,40 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+
+  <body class="page">
```
Sveta-Kryukova, line 10: don't we need to do a separate BEM block for it?

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    @include square(100px);
+
+    text-align: center;
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
```
Sveta-Kryukova: maybe we can move it to normalize block

src/styles/main.scss
```diff
@@ -1,3 +1,6 @@
-body {
-  margin: 0;
-}
+@import "./utils/normalize.scss";
+@import "./utils/variables.scss";
+@import "./utils/functions.scss";
+@import "./utils/mixins.scss";
+
```
Sveta-Kryukova, line 6: also we need a reset block=)

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,14 @@
...
+
+$days-name: (
+  mon:1,
+  tue:2,
+  wed:3,
+  thu:4,
+  fri:5,
+  sat:6,
+  sun:7,
+);
```
Sveta-Kryukova: I think that it is not a mistake but in video was another one, but as I told I don't sure that this is mistake)

Sveta-Kryukova (changes requested):

src/styles/main.scss
Anastasia4sik: As I understood it`s not necessary, but I`ll add it, thank you.

src/styles/main.scss
```diff
@@ -1,3 +1,8 @@
-body {
-  margin: 0;
-}
+@import "utils/normalize.scss";
+@import "utils/variables.scss";
+@import "utils/functions.scss";
+@import "utils/mixins.scss";
+@import "utils/reset.scss";
```
Sveta-Kryukova, line 5: we should firstly import reset.scss and than normalize.scss

Sveta-Kryukova (approved):

solaryasha (approved):
Good job!

PR #1415, opened 2022-11-09
calendar

grslckr:
- [DEMO LINK](https://grslckr.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,48 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendarWidth;
+  gap: $calendarGap;
+  padding-top: $calendarSide;
+
+  @each $number, $day in $days {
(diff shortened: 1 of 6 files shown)


lukacsja (approved):
nothing to say - i guess
nice one

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendarWidth;
+  gap: $calendarGap;
+  padding-top: $calendarSide;
```
meforoma, line 6: not that it impacts anything, but just wondering why would you need it?
without padding it looks more centered.
But of course it's a matter of preference)

meforoma (approved):
👍

src/styles/blocks/calendar.scss
grslckr: yes, i don't actually need it but the bottom space looked ugly to me that's why i put it 😁

PR #1416, opened 2022-11-09
add calendar file to git

batstolya:
- [DEMO LINK](https://batstolya.github.io/layout_calendar/)

### DIFF
--- a/src/styles/block/calendar.scss
+++ b/src/styles/block/calendar.scss
@@ -1,0 +1,60 @@
+.calendar {
+  padding: 10px;
+  margin: 0 auto;
+  width: ($width-block-calendar-day * $days-in-row)
+    + ($gap-calendar * $days-in-row);
+
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap-calendar;
+  @for $i from 28 through 31 {
+    &--month-length--#{$i}
+    &__day:nth-child(n + #{$i + 1}) {
+      display: none;
+    }
+  }
+
+  @for $i from 1 through $count-of-days {
+    &__day:nth-child(n + #{$i})::before {
+      content: "#{$i}";
+    }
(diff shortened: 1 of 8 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,17 @@
...
+$width-container-all: $width-block-calendar-day * $days-in-row + $gap-calendar * $days-in-row;
+
+
+$color-regular: #060b35;
+$color--button--blue: #00acdc;
+$color--main--white: #fff;
+
+$card-width: 200px;
+$star-size: 16px;
+$height-button: 40px;
```
OlggaJul: completely reduntant variables from previous task))

src/styles/block/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+    color: #000;
+    content: "";
+    font-weight: arial;
+    font-size: 30px;
+
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+  }
```
OlggaJul: you can move this styles inside .calendar__day and use flex for positioning. It would be easier but it's up to you) 
font-weight: arial ??

OlggaJul (changes requested):

src/styles/utils/_variables.scss
batstolya: @OlggaJul ha-ha 
did not notice) thanks

src/styles/block/calendar.scss
batstolya: @OlggaJul yes, that's a good idea, thanks, I'll fix it,
 thanks

ivan-kadykalo (approved):
🔥

PR #1422, opened 2022-11-09
Calendar task done!

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/layout_calendar/)


.linthtmlrc.json
```diff
@@ -1,3 +1,49 @@
 {
-  "extends": "@mate-academy/linthtml-config"
+  "attr-bans": [
```
UmizDemud, line 2: Pushing the **.linthtmlrc.json** file is generally not good. Because, the changes in this file  are generally only concerns you. 
Adsız
Adsız2

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day calendar--month-length">
```
UmizDemud: The class here should be, 

calendar--start-day-wed

You see, with a  **-wed**  at the end

so that the calendar knows where to start.

UmizDemud: Also this class should be,

calendar--start-day-29

With a -29 at the end

so the calendar knows how many days it should go.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,31 @@
...
+    "mon" 1,
+    "tue" 2,
+    "wed" 3,
+    "thu" 4,
+    "fri" 5,
+    "sat" 6,
+    "sun" 7;
+
+  @each $name, $value in $days {
+    &--start-day-#{$name}:first-child {
```
UmizDemud:
```scss
    &--start-day-#{$name} :first-child {

```

There must be a space before :first-child , otherwise it will not look inside the selector but it will find the first div with a class .calendar--start-day-29 in the parent.. Try and see

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,31 @@
...
+      margin-left: ($value - 1) * ($cellWidth + 1px);
+    }
+  }
+
+  @for $var from 28 through 31 {
```
UmizDemud, line 26: There needs to be one more for loop inside to hide all days after 29
And just like before there needs to be a space before the ```:first-child``` for the same reason with the above comment.


```scss
  @for $var from 28 through 31 {
    @for $i from 1 through 31 - $var { // This code goes through 1,2,3... hiding the elements after the 29th day
      &--month-length-#{$var} :nth-child(#{$var + $i}) {  // day 29 is an example I'm using.. Works for 28,29,30
        display: none;
      }
    }
  }
```

UmizDemud (changes requested):
The code looks clean, neatly organized. Please take a look at the notes I left, and see if you can work it out. Otherwise you can text me for any question.

src/styles/blocks/calendar.scss
mmahfuzi: > There needs to be one more for loop inside to hide all days after 29 And just like before there needs to be a space before the `:first-child` for the same reason with the above comment.
> 
> ```css-scss
>   @for $var from 28 through 31 {
>     @for $i from 1 through 31 - $var { // This code goes through 1,2,3... hiding the elements after the 29th day
>       &--month-length-#{$var} :nth-child(#{$var + $i}) {  // day 29 is an example I'm using.. Works for 28,29,30
>         display: none;
>       }
>     }
>   }
> ```

src/index.html
mmahfuzi: > Also this class should be,
> 
> calendar--start-day-29
> 
> With a -29 at the end
> 
> so the calendar knows how many days it should go.

I didn't get this one, why shall I add -29 at the end and should there be another calendar--start-day-? like this:
calendar--start-day-wed    calendar--start-day-29

mmahfuzi:
> The code looks clean, neatly organized. Please take a look at the notes I left, and see if you can work it out. Otherwise you can text me for any question.

Thanks for all your comments, I applied them :).

UmizDemud (approved):
Great job!

meforoma (approved):
🎉 

@UmizDemud very thorough review 👍

PR #1427, opened 2022-11-10
initial commit _ calendar was created

DimaKolesnyk:
- [DEMO LINK](https://dimakolesnyk.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
+
+.calendar {
+  position: relative;
+  top: 50%;
+  left: 50%;
```
viktoriia-kravchenko: Let's move these styles to page__calendar class.

src/styles/main.scss
```diff
@@ -1,3 +1,7 @@
-body {
-  margin: 0;
-}
+@import "./utils/variables";
+@import "./utils/reset";
+@import "./utils/normalize";
```
viktoriia-kravchenko, line 3: We can first mention reset and normalize, and after that variables.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+  transform: translate(-50%, -50%);
+
+
+  display: flex;
+  flex-wrap: wrap;
+  gap: $dayBoxGap;
+  width: getCalendarWidth($dayBoxSize, $calendarPadding, $dayBoxGap);
+  padding: 10px;
+
```
viktoriia-kravchenko: I propose to use $calendarPadding here and increase vertical paddings, since the top row rises to a height of 20px on hover.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+      &:nth-child(#{$date})::before {
+        content: "#{$date}";
+      }
+    }
+
+    @each $day, $margin in $days {
+      .calendar--start-day--#{$day} &:first-child {
+        margin-left: $dayBoxSize * $margin + $margin;
+      }
+    }
```
viktoriia-kravchenko, line 39: Let's move it higher to calendar styles because it's a modifier for the calendar.

viktoriia-kravchenko (changes requested):
Good job! 
Please, add @for loop to set the last day to show (use modifier month-length and nth-child).

src/styles/blocks/calendar.scss
DimaKolesnyk: It's a separate BEM block and in order to be reusable it can't be a page element

src/styles/main.scss
DimaKolesnyk: What to do in case if I need to use variable inside of normalize.scss? It simply won't be readable.

src/styles/blocks/calendar.scss
DimaKolesnyk: Thx for advice

src/styles/blocks/calendar.scss
DimaKolesnyk: It's not working  in that way

viktoriia-kravchenko (approved):

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
+
+.calendar {
+  position: relative;
+  top: 50%;
+  left: 50%;
```
valerii-derkach: yes it is, but @viktoriia-kravchenko is right, these styles are external position of calendar, so use mix here

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,13 @@
...
+$days:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
+
+//////
```
valerii-derkach: remove it

valerii-derkach (changes requested):
fix the month-length modifier, it doesn't work

Знімок екрана 2022-11-15 о 09 47 49

src/styles/blocks/calendar.scss
DimaKolesnyk: I got it, thx

valerii-derkach (approved):

PR #1447, opened 2022-11-10
complete task

Alireza-Mohammadhossein:
- [DEMO LINK](https://Alireza-Mohammadhossein.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+      margin-left: $index * ($day-width-height + $day-gap);
+    }
+  }
+
+  @for $last-day from 28 through 31 {
+    &--month-length-#{$last-day} :nth-child(n + #{$last-day + 1}) {
+      display: none;
+    }
+  }
+}
```
mmahfuzi, line 60: There needs to be one more for loop inside to hide all days after 29.

mmahfuzi:
```suggestion
  @for $last-day from 28 through 31 {
    @for $i from 1 through 31 - $last-day { // This code goes through 1,2,3... hiding the elements after the 29th day
      &--month-length-#{$last-day} :nth-child(#{$last-day + $i}) {  // day 29 is an example I'm using.. Works for 28,29,30
        display: none;
      }
    }
  }
```

mmahfuzi (changes requested):
I only found one change to be made (adding a for loop) in order for your calendar to work as expected. The rest of your code looks nice. :)

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,6 @@
+@use "../utils/mixin" as *;
+
+.page {
+  @include center;
```
IoghiStudio, line 4: You can wrap the calendar or make it an element of the page and use margin: 0 auto to center just the calendar

src/styles/blocks/calendar.scss
Alireza-Mohammadhossein: @mmahfuzi thank you for your review, but I checked it, and it works with my loop. when I change the Html class to "calendar--month-length-29" or "calendar--month-length-30" it works

src/index.html
```diff
@@ -4,9 +4,44 @@
...
-    <link rel="stylesheet" href="./styles/main.scss">
+    <!-- linthtml-configure attr-bans=false -->
+    <link rel="stylesheet" href="./styles/main.css">
   </head>
-  <body>
-    <h1>Calendar</h1>
+
+  <body class="page">
+    <div class=
+      "calendar calendar--start-day-wed calendar--month-length-30">
```
naziamoff, line 13: You should count each classname as a separate attribute if you have more than 2 of them
So in this case better to make each of them into a separate line
```suggestion
    <div class="
      calendar 
      calendar--start-day-wed 
      calendar--month-length-30
    ">
```

naziamoff (approved):
Well done 🔥 
I did really like the transition when I change the days quantity with dev tools 👍

PR #1455, opened 2022-11-11
-codeupdated-

Grayfox147:
- [DEMO LINK](https://Grayfox147.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Grayfox147.github.io/layout_calendar/report/html_report/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px ;
+  padding: 25px;
+
```
alaaashkar: no need for these spaces

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  margin: auto;
+  width: (7 * $cell-width) + 10px;
+  // Use flex with 1px gap
+  // and limit its width to
+  // exactly 7 columns + 10px paddings
+
+  $days:
```
alaaashkar: better move this map to the variable files

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+  background-color: #eee;
+  border: 1px solid black;
+  font-size: 30px;
+  text-align: center;
+  transition: 0.5s;
+
+  &:hover {
+    background-color: pink;
+    transform: translateY(-20px);
+
```
alaaashkar: no need for this space

src/styles/main.scss
```diff
@@ -1,3 +1,8 @@
+@import "./utils/variables.scss";
+
+@import "./blocks/calendar.scss";
+
 body {
```
alaaashkar: better add the font family property and anything related to overall project to the 'html' tag because it has to be applied everywhere not only for the body for example header and footer tags and to do that : create a new file in the 'utils' folder and call it 'normalize' for example and move these properties there

alaaashkar: also move the body property to a single file in utils and call it 'resets' for example to reset the margins and paddings and leave the main file only for importing the scss files without any extra thing inside

alaaashkar (changes requested):
just simple simple changes and boom it is ready to go ! Amazing work overall wow !

src/styles/blocks/calendar.scss
Grayfox147: Done!

src/styles/blocks/calendar.scss
Grayfox147: DOne!

src/styles/blocks/calendar.scss
Grayfox147: Done!

src/styles/main.scss
Grayfox147: Done!

Grayfox147: Done!

alaaashkar (approved):
looks great to me ! amazing !

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+    background-color: pink;
+    transform: translateY(-20px);
+    cursor: pointer;
+  }
+
+  @for $day from 1 through 31 {
+    &--day#{$day}::before {
+      content: "#{$day}";
+      line-height: $cell-height;
+    }
```
meforoma: you could simplify it quite a bit
```
    @for $day from 1 through 31 {
      &:nth-child(#{$day})::before {
        content: "#{$day}";
				...
      }
    }
```

and then you won't need the whole class `calendar__day--day1...31`

let's make it work @Grayfox147

src/index.html
```diff
@@ -1,12 +1,44 @@
...
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-wed calendar-month-length--29">
+      <div class="calendar__day calendar__day--day1"></div>
```
meforoma: remove seconds class for all days
to better understand the idea - check the comment in css
```suggestion
      <div class="calendar__day"></div>
```

meforoma (changes requested):
good job 👍 
let's optimise the code, and you score it 🎯

src/styles/blocks/calendar.scss
Grayfox147: Ahhh I see! if I use the nth-child $day I can replace the code from HTML

src/index.html
Grayfox147: All done!

jackelashkar (approved):
good job!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+  border: 1px solid black;
+  font-size: 30px;
+  text-align: center;
+  transition: 0.5s;
+
+  &:hover {
+    background-color: pink;
+    transform: translateY(-20px);
+    cursor: pointer;
+  }
```
naziamoff, line 37: that is scss so you can just write it inside the `.calendar` block as `&__day`

naziamoff (approved):
Well done 🔥 
Just one small thing

meforoma (approved):

PR #1467, opened 2022-11-12
Develop

IoghiStudio:
- [DEMO LINK](https://IoghiStudio.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,70 @@
...
+
+      background: #eee;
+      border: 1px solid #000;
+      cursor: pointer;
+      transition-duration: 350ms , 700ms ;
+      transition-property: transform, background-color ;
+
+      &:hover {
+        transition: transform 700ms;
+        transform: translate(10px , 15px);
```
Grayfox147, line 31: Here I would recommend you to use transform: translateY(-20px); so basically the movement that they want is uppwards so you use the Y to move in the Axis vertical and if you use positive it would move downwards you want it up so you use  a negative value.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,70 @@
...
+        z-index: 2;
+        background-color: #ca71db;
```
Grayfox147, line 33: I believe the pink they are looking for is more light pink! you could try with a different one or just try "pink"

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,70 @@
...
+      content: "#{$day}";
+      color: #000;
+    }
+
+  }
+
+  $startDay:
+    "mon" 0,
+    "tue" $daySquareSize + $calendarGap,
+    "wed" $daySquareSize * 2 + $calendarGap * 2,
```
Grayfox147, line 47: This information you could store it in variables. with the rest of your variables!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,70 @@
...
+
+      box-sizing: border-box;
+      width: $daySquareSize;
+      height: $daySquareSize;
+
+      background: #eee;
+      border: 1px solid #000;
+      cursor: pointer;
+      transition-duration: 350ms , 700ms ;
+      transition-property: transform, background-color ;
```
Grayfox147, line 27: since you are using transition duration same value more than once you could create a variable and store it in the variable.scss and replace it here.

Grayfox147 (changes requested):
I believe overall you did a good job, different approach than mine in the For Loops or the each loop but really good job!

Grayfox147:
Also remember to check the box called checklist in the excel file so the mentors realize you already checked it before pushing the result! otherwise they wont approve.

src/styles/blocks/calendar.scss
IoghiStudio: Was an optional task so is not wrong if is other animation

src/styles/blocks/calendar.scss
IoghiStudio: Was an optional task.

Grayfox147 (approved):

src/index.html
```diff
@@ -1,12 +1,50 @@
...
+        initial-scale=1.0
+      ">
     <title>Calendar</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="stylesheet" href="./styles/index.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page__content">
+    <!-- Is also an element of page so we can apply margin -->
```
naziamoff, line 15: you should leave comments in your code only if they are very very needed

src/index.html
```diff
@@ -1,12 +1,50 @@
...
+    <div class="calendar calendar--month_length-29 calendar--start_day-fri page__calendar">
+      <div class="calendar__day-1"></div>
+      <div class="calendar__day-2"></div>
+      <div class="calendar__day-3"></div>
+      <div class="calendar__day-4"></div>
+      <div class="calendar__day-5"></div>
+      <div class="calendar__day-6"></div>
```
naziamoff, line 22: think of a way of making loop that helps you to avoid writing `...day-x` explicitly
You don't want anything to be hardcoded in your code

src/styles/blocks/calendar.scss
naziamoff: @Grayfox147 agree 👍

src/styles/blocks/calendar.scss
naziamoff: @IoghiStudio I like your hover but now it looks a bit counterintuitive (me as user expect that it will go up on hover but not down). So I would recommend you to change the direction of your translate

naziamoff (approved):
LGTM overall 🔥 
But consider my comments in your next tasks

src/index.html
IoghiStudio: I did that when i started the project and forgot to delete it after. I understand!

src/index.html
IoghiStudio: That's true , ill have to think more next time!

PR #1498, opened 2022-11-23
Calendar

bojkovladislav:
- [DEMO LINK](https://bojkovladislav.github.io/layout_calendar/)

### DIFF
--- a/src/styles/utils/_colors.scss
+++ b/src/styles/utils/_colors.scss
@@ -1,0 +1,3 @@
+$days-color:#eee;
+$font-color: #000;
+$day-hover-color:pink;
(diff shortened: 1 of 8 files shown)


src/styles/utils/_colors.scss
```diff
@@ -0,0 +1,3 @@
+$days-color:#eee;
```
Viktor-Kost, line 1: Better use some abstract names for variables
Perhaps you will need to use this color not only for days
You can use something like 
$primary-color
$secondary-color

Viktor-Kost (approved):
Good job!

Just apply one recommendation

PR #1525, opened 2022-12-06
Added calendar

annperetiatko:
- DEMO LINK
https://annperetiatko.github.io/layout_calendar/
- TEST REPORT LINK
https://annperetiatko.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,46 @@
+.calendar {
+  box-sizing: border-box;
+  background-color: #eee;
+  border: 1px solid #000;
+  transition: 0.5s;
+  @include blockSize;
+
+  &:hover {
+    background-color: #ffc0cb;
+    transform: translate(0, -20px);
+    cursor: pointer;
+  }
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  box-sizing: border-box;
+  background-color: #eee;
+  border: 1px solid #000;
+  transition: 0.5s;
+  @include blockSize;
```
witflash, line 6: It's better to separate mixins and directives from other rules with an empty line.

src/styles/main.scss
```diff
@@ -1,3 +1,12 @@
+@import "./utils/variables";
+@import "./utils/mixins";
+@import "./blocks/calendar";
+
 body {
```
witflash, line 5: It's better to add an additional wrapper and pass these rules for it instead of directly `body`. Not critical for this project but a good point for the future work.

witflash (approved):
Good job!

PR #1542, opened 2022-12-08
Calendar was created

andrieieva-yuliia:
- [DEMO LINK](https://andrieieva-yuliia.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -7,6 +7,7 @@
   gap: $gap-size;
   width: 7 * $day-size + 6 * $gap-size;
   padding: 10px;
+  transition: transform $animation-duration;
 
   @for $day from 28 through 30 {
     &--month-length--#{$day} :nth-child(n+#{$day+1}) {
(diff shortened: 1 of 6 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+    "sun" 6;
+
+  @each $start-day, $start-day-value in $days-of-week {
+    &--start-day--#{$start-day} :first-child {
+      margin-left: $start-day-value * ($day-size + $gap-size);
+    }
+  }
+}
+
+.calendar__cell {
```
anna-inozemets: Use SASS nesting. It should look like:
.calendar [your main BEM block { 
    &__cell [main BEM block's childern] {
    }
}

Checklist: checkpoint 4

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+      content: "#{$day-number}";
+      margin: auto;
+    }
+  }
+
+  &:hover {
+    cursor: pointer;
+    transform: translateY(-20px);
+    animation-duration: $animation-duration;
```
anna-inozemets: I am not sure, but for this better use property "transition" and don't put it under the hover pseudoclass, just add it for calendar__cell, like 
transition: transform $animation-suration

anna-inozemets (changes requested):
For sure, your calendar is pretty nice. Good job!

src/styles/blocks/calendar.scss
andrieieva-yuliia: I totally agree with you, I missed this moment, thanks 🤩

src/styles/blocks/calendar.scss
andrieieva-yuliia: I agree about the `transition`, but I need to leave the `transform: translateY(-20px);` in the hover to set the direction) thanks a lot 🌷

ArturErmolenko (approved):

PR #1568, opened 2022-12-13
Added calendar

burovetskyi:
- [DEMO LINK](https://burovetskyi.github.io/layout_calendar/)


src/styles/utils/_variables.scss
```diff
@@ -1,4 +1,13 @@
-$days: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
-$cell-side: 100px;
-$gap: 1px;
-$calendar-padding: 10px;
+$container-day-width: 100px;
+$container-day-height: 100px;
+$day-gap: 1px;
+$calendar-width: 7 * $container-day-height + 20px + 6 * $day-gap;
```
Coyoote: Here you should multiply 7 * $container-day-width instead, because if the width of the day would be changed, your calendar width will stay the same.

src/styles/utils/reset.scss
```diff
@@ -4,4 +4,9 @@
 
 body {
   margin: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
```
Coyoote: This shouldn't be in the reset file, as you're assigning properties instead of resetting them. You should add this to the container that holds your calendar like your  `.container` for example, and create a separate file for that BEM block

src/styles/blocks/calendar.scss
```diff
@@ -1,45 +1,43 @@
...
+  padding: 70px 10px;
 
   &__day {
     display: flex;
     justify-content: center;
     align-items: center;
-    @include squareCell($cell-side);
-    background: #eee;
+    width: #{$container-day-width};
+    height: #{$container-day-height};
```
Coyoote: You can write your variables without the #{variable-name}, you can use $variable-name instead.

Coyoote (commented):
Thanks for letting me review your work. You did great, although some changes need to be implemented. Hope I was of good help 😅

alexandra-protyanova (approved):
Good job 👍

PR #1579, opened 2022-12-13
calendar task

stadnytskyi-ivan:
- [DEMO LINK](https://stadnytskyi-ivan.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,41 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  padding-top: $calendar-top-padding;
+  gap: $day-gap;
+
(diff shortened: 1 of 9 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,41 @@
...
+  gap: $day-gap;
+
+  &__day {
+    @include square ($day-size);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 1px solid black;
+    background-color: $day-color;
+    transition: 0.5s all;
```
vladyslav78292, line 15: good job! I did not find any mistakes, but I think it would be better to add a transition only for color and transform properties, instead of all properties

denys-cheporniuk (approved):
good job 🔥

PR #1580, opened 2022-12-13
Calendar

IvanLugovskiy:
- [DEMO LINK](https://IvanLugovskiy.github.io/layout_calendar/)
- [TEST REPORT LINK](https://IvanLugovskiy.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
+@import "./variables.scss";
+
 body {
   margin: 0;
 }
+
+* {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
```
polosanya: Just set the font for the whole page and inherit where needed. No need to use * selector here

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  gap: 1px;
+  margin-top: 100px;
+  margin-left: $general-margin;
+  margin-right: $general-margin;
```
polosanya: You can use shorthand property
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#syntax

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  text-align: center;
+  
+  &__elem {
+    display: block;
+    width: 98px;
+    height: 98px;
+    background-color: $color-main;
+    border: 1px solid $color-border;
+    transition: 500ms;
```
polosanya: Better to create a variable for this value. Also, please specify to which exactly property you want to apply transition because now it applies to all the properties and it's really bad for performance

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  gap: 1px;
+  margin-top: 100px;
+  margin-left: $general-margin;
+  margin-right: $general-margin;
+  text-align: center;
+  
+  &__elem {
+    display: block;
+    width: 98px;
+    height: 98px;
```
polosanya: Think about changing box-sizing, so you don't need to set sizes without borders. Also, would be great to create a variable for elem size.
https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+    background-color: $color-main;
+    border: 1px solid $color-border;
+    transition: 500ms;
+  }
+
+  &__elem:hover {
+    transform: translateY(-20px);
+    transition-duration: 500ms;
```
polosanya: You've already set it above

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+    "tue" $two,
+    "wed" $three,
+    "thu" $four,
+    "fri" $five,
+    "sat" $six,
+    "sun" $seven;
+
+  @each $wday, $num in $weekdays {
+    &--start-day-#{$wday} {
+      .calendar__elem:nth-child(1) {
```
polosanya: Try this:
https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child

polosanya (changes requested):
Looks great! Let's just improve your code a bit)

src/styles/main.scss
IvanLugovskiy: last home work i had changed box-sizing values and i had tip from mentor that it box-sizing better to set value 'content-box' ofcourse i will change if it is not correct but, could you explaine to me how it s correct in generaly ?

src/styles/main.scss
IvanLugovskiy: sure)

src/styles/main.scss
loralevitska: `box-sizing: border-box` is better for this task. In this way, you will not need to take into account the size of the borders and calculate them yourself from the size of the content.
Regarding the previous case, if you do not understand, then ask in the fe_chat and tag the mentor so that they explain to you why this particular solution is better for that task

loralevitska (approved):
good job!

PR #1635, opened 2022-12-29
calendar develop

Anatolii-Polishchuk:
https://Anatolii-Polishchuk.github.io/layout_calendar/
https://Anatolii-Polishchuk.github.io/layout_calendar/report/html_report/


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    box-sizing: border-box;
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
+    @include square($box-size);
+    &:hover {
+      background-color: #ffc0cb;
+      position: relative;
+      bottom: 20px;
+      cursor: pointer;
+      transition-duration: 500ms;
```
qoa11a: Let's add transition rules not on hover, but on whole card. And also you need to specify to which properties this transition-duration is applied.

qoa11a (changes requested):

SanyaBratashchuk (approved):

PR #1640, opened 2023-01-02
the task calendar completed

MaksymHuriev:
- [DEMO LINK](https://MaksymHuriev.github.io/layout_calendar/)
- [TEST REPORT LINK](https://MaksymHuriev.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -3,7 +3,6 @@
   flex-wrap: wrap;
   gap: $gap-day;
   width: $calendar-width;
-  padding: 100px 0;
 
   &__day {
     display: flex;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,36 @@
...
+      }
+    }
+  }
+
+  @each $day, $number in $calendar-week {
+    &--start-day--#{$day} &__day:first-child {
+      margin-left: $number * $width-day + $number * $gap-day;
+    }
+  }
+}
```
DarkMistyRoom: add an empty line in the end

DarkMistyRoom (changes requested):
- The calendar goes over the left side without scroll. That's because of flex on the `container` try to center the calendar in another way

image

- Length modifier doesn't work
image

DarkMistyRoom (changes requested):
this problem is still here. that's because of flex on the container. try to find a solution. If you have any problems with this, write in the chat
image

agniya-i (approved):

PR #1647, opened 2023-01-04
develop

vova552:
- [DEMO LINK](https://vova552.github.io/layout_calendar/)
- [TEST REPORT LINK](https://vova552.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+$the-last-day: 31;
+$feb-day-count: 28;
+$week:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
```
Lilu580: for variables better create new file _variables.css

src/index.html
```diff
@@ -7,6 +7,41 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
```
Lilu580: for what you crated this container?

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
```
Lilu580: same question for .container

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+  gap: $calendar-day-gap;
+  width: 7 * $calendar-day-width + 16px;
+
+  &__day {
+    box-sizing: border-box;
+    height: $calendar-day-width;
+    width: $calendar-day-width;
+    border: 1px solid black;
+    background-color: #eee;
+    transition: transform 0.3s;
```
Lilu580: in description for task The hovered cell has to become pink says:

    (Optional) Move it up by 20px (use transform)
    (Optional) both of these properties are to be animated with the duration of half a second.Both properties and 0.5w

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+  align-items: center;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-day-gap;
+  width: 7 * $calendar-day-width + 16px;
+
```
Lilu580: here you need to set vertical padding: checklist 11.[STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.

and from where you get 16px? and you forgot about gaps.Your with should be 
```suggestion
 max-width:($calendar-day-width * $days-in-raw) + ($calendar-day-gap * ( $days-in-raw - 1)) 
```You can realize this with function

File: src/styles/main.scss [hunk 6]
```diff
@@ -1,3 +1,69 @@
...
+        justify-content: center;
+        align-items: center;
+        height: 100%;
+      }
+    }
+  }
+
+  @each $cell, $number in $week {
+    &--start-cell_#{$cell} &__day:first-child {
+      margin-left: ($calendar-day-width + $calendar-day-gap) * $number;
```
Lilu580, line 52: You can realize this with func if you want

Lilu580 (commented):
Good

src/index.html
vova552: it's easier that way

src/styles/main.scss
vova552: it becomes pink

src/styles/main.scss
vova552: but it works like that too

src/styles/main.scss
andriibezkorovainyi: You could add mix class to calendar and add postition styles to it. This way we'll not gonna add any extra blocks to our page and more accurately follow the task.

src/styles/main.scss
andriibezkorovainyi: > but it works like that too

I agree, you don't need to add exactly paddings to center your calendar. But the calculation of width could be more clear:

- Use properly named variables to make all the calculations more clear.

It is best to use function but not necessary. Just make your width calculation include more variables. This solution is good: 
`
 max-width:($calendar-day-width * $days-in-raw) + ($calendar-day-gap * ( $days-in-raw - 1)) ;
`

src/index.html
andriibezkorovainyi: > 

Its better to add mix to our calendar rather then create new blocks

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-day-gap;
+  width: 7 * $calendar-day-width + 16px;
+
+  &__day {
+    box-sizing: border-box;
+    height: $calendar-day-width;
```
andriibezkorovainyi, line 27: You can add properties such as box-sizing to normalize file or in the start of the main.scss using 

`* {
  box-sizing: border-box;
}`

src/styles/main.scss
andriibezkorovainyi: > it becomes pink

Yes, it does but it becomes pink immediately.
8. [STYLES] - Make sure to list all styles that you apply transition to.

You need to add color to transition

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+
+    @for $index from 1 through $the-last-day {
+      &:nth-child(#{$index})::before {
+        content: "#{$index}";
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        height: 100%;
+      }
+    }
```
andriibezkorovainyi, line 47: 7. [SASS] - Don't use SASS loops for styles that stay the same for all elements of the group, e.g. display or position.

andriibezkorovainyi (changes requested):
Great job! I've found some little issus with your code, you can solve it and make your project really clean.

andriibezkorovainyi (approved):
👍👍

src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
...
+
+@import "./variables";
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
+
+.calendar {
```
sergii-nosachenko, line 13: Checklist #3

src/index.html
```diff
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--container calendar--start-cell_sun calendar--month-length_29">
```
sergii-nosachenko, line 10: Overloaded. You don't need `calendar--container`.
Also read task carefully about modifier names.

src/styles/_variables.scss
```diff
@@ -0,0 +1,13 @@
+$calendar-day-width: 100px;
+$calendar-day-gap: 1px;
+$the-last-day: 31;
+$feb-day-count: 28;
+$days-in-raw: 7;
```
sergii-nosachenko, line 5:
```suggestion
$days-in-row: 7;
```

src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
+* {
```
sergii-nosachenko, line 1: It's a reset/normalize, `body` rule too. Move to separate file.

src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
...
+  font-size: 30px;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-day-gap;
+  width: $days-in-raw * $calendar-day-width + 16px;
+
+  &--container {
```
sergii-nosachenko, line 19: Move to `.calendar`

src/styles/main.scss
sergii-nosachenko: Fix student comment. Checklist #7

src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
...
+    left: 25%;
+  }
+
+  &__day {
+    box-sizing: border-box;
+    height: $calendar-day-width;
+    width: $calendar-day-width;
+    border: 1px solid black;
+    background-color: #eee;
+    transition: background-color, transform 0.3s;
```
sergii-nosachenko, line 31: Wrong duration. Also, no value for `background-color` - it changes immediately.

sergii-nosachenko (changes requested):

PR #1648, opened 2023-01-04
Calendar

SashaKozoriz:
- [DEMO LINK](https://SashaKozoriz.github.io/layout_calendar/)
- [TEST REPORT LINK](https://SashaKozoriz.github.io/layout_calendar/report/html_report/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+      }
+    }
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
```
Viktor-Kost, line 40: move `cursor: pointer;` to the static state (&__cell)
```suggestion
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
+.calendar {
+  margin: auto;
+  padding: 50px $hor-padding $hor-padding;
```
Viktor-Kost, line 3: it's not obvious to use variables like this, because it's not unified. 
Why we have $hor-padding and don't have vertical one? 
I would use paddings here without variables

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+  max-width: ($cell-size*$cell-per-row)+
+    ($gap*$cell-per-row - $gap)+
+    ($hor-padding*2);
```
Viktor-Kost, line 6: it's better to use function here instead of such calculation here
If you haven't worked with sass functions - feel free to ask at fe_chat

Viktor-Kost (changes requested):
Good job!!

Need some improvements

PR #1651, opened 2023-01-05
Calendar grid created

kateryna-tilishevska:
- [DEMO LINK](https://kateryna-tilishevska.github.io/layout_calendar/)
- [TEST REPORT LINK](https://kateryna-tilishevska.github.io/layout_calendar/report/html_report/)


src/styles/utils/normalize.scss
```diff
@@ -0,0 +1,12 @@
+* {
+  box-sizing: border-box;
+}
+
+li,
+p,
+ul,
+body {
+  list-style: none;
```
witflash: You don't need this rule for every these tags but just for `ul`. So create separate selector for `ul` with this rule.

witflash (changes requested):
Good work! Just do small improvement with my comment.

src/styles/utils/normalize.scss
kateryna-tilishevska: Thank you!

IvanFesenko (approved):

PR #1657, opened 2023-01-09
tasks solved

DanielLinker17:
- [DEMO LINK](https://DanielLinker17.github.io/layout_calendar/)
- [TEST REPORT LINK](https://DanielLinker17.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,59 @@
+body {
+  margin: 0;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $day-gap;
+  max-width: $week-length;
+  position:absolute;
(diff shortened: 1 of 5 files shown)


SanyaBratashchuk (changes requested):
Doesn't work as expected. Amount of days and first day of the month should change when I change modifier's value 
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
+.page {
+  margin: 0;
+}
```
agniya-i: it's would be better to move this into `main.scss`, and keep only those styles that related to the calendar in this file

agniya-i (changes requested):

SanyaBratashchuk (approved):

PR #1660, opened 2023-01-11
calendar is created

vanyachyzh:
- [DEMO LINK](https://vanyachyzh.github.io/layout_calendar/)
- [TEST REPORT LINK](https://vanyachyzh.github.io/layout_calendar/report/html_report/)

I don't know how to move variables to _variables.scss. If I do this my path become broken. Maybe it is result of using @import url() insted of @use. So I decided to leave them in their local files so far


qoa11a (changes requested):
image
According to the task if modifier is out of range there should be 31 days.

witflash (approved):
Well done 👍

PR #1672, opened 2023-01-12
Layout Calendar was created

vlad-ptashnichenko:
[DEMO LINK](https://vlad-ptashnichenko.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,45 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: 710px;
+  margin: 20px;
+  gap: 1px;
+
+  &__day {
+    width: $day-block-width;
+    box-sizing: border-box;
+    font-family: Arial, Helvetica, sans-serif;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: 710px;
+  margin: 20px;
+  gap: 1px;
+
+  &__day {
+    width: $day-block-width;
+    box-sizing: border-box;
```
witflash, line 10: You can use this rule for `*` selector at any project by default. 
But here it's ok.

src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
-body {
-  margin: 0;
-}
+@import "./utils/extend";
+@import "./utils/variables";
+
+@import "./blocks/reset.scss";
```
witflash, line 4: `reset` stylesheet should be the first one. And after it all your own styles.

witflash (approved):
Well done 👍

PR #1676, opened 2023-01-13
Develop

yarik-bachynskyi:


src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
 body {
   margin: 0;
 }
+
+$time: 0.6s;
+$wite: #eee;
+$black: #000;
+$pink: #ffb6c1;
```
ViktorHavryliuk: It's better to give variable names based on their function, not exact color, for example "hover-color" or "secondary-color", etc. This way if in future it is desided to change pink to some other color, you won't have to change variable name every time

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
 body {
   margin: 0;
 }
+
+$time: 0.6s;
```
ViktorHavryliuk: Better be more specific with names, if it is not too long. Here "transition-time" would be more representative of what this variable exists for

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+  wed: 2,
+  thu: 3,
+  fri: 4,
+  sat: 5,
+  sun: 6,
+);
+
+.calendar {
+  width: 760px;
+  margin: 0 auto;
```
ViktorHavryliuk: Don't apply margins to BEM blocks, better use bem mix or wrapper instead

ViktorHavryliuk (changes requested):
Add some padding on top of the page, to prevent first row of cells fron hovering behind it:
image

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+    &--start-day--#{$day} > :first-child {
+      margin-left: $order * (100px + 1px);
+    }
+  }
+
+  &__cell {
+    display: flex;
+    margin-top: 1px;
+    box-sizing: border-box;
+    width: 100px;
```
ViktorHavryliuk, line 44: It's better to move dimensions, like width and height, to vars too, as well as gap between cells

ViktorHavryliuk (approved):
Approved, but please also move cells size and gap between them into variables

PR #1703, opened 2023-01-23
Calendar Done

stanislavcodes:
https://stanislavcodes.github.io/layout_calendar/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -14,7 +14,7 @@
 
   <body>
     <main class="main">
-      <div class="calendar main__calendar calendar--starting-day-thu calendar--month-length-30">
+      <div class="calendar main__calendar calendar--starting-day--thu calendar--month-length--30">
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -6,39 +6,70 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
     <main class="main">
       <div class="main__calendar calendar calendar--starting-day-thu calendar--month-length-30">
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
         <div class="calendar__day"></div>
+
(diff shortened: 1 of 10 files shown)


src/index.html
```diff
@@ -6,7 +6,72 @@
...
-    <h1>Calendar</h1>
+    <main class="main">
+      <div class="main__calendar calendar calendar--starting-day-thu calendar--month-length-30">
+        <div class="calendar__day"></div>
+
+        <div class="calendar__day"></div>
+
+        <div class="calendar__day"></div>
+
+        <div class="calendar__day"></div>
```
kuznetsova-anastasiia: i don't think it has to be empty lines between each tag in this case, but it's on your taste mb

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    &:hover {
+      background-color: #fcbfcb;
+      transform: translateY(-20px);
+    }
+
+    @for $day-number from 1 through 31 {
+      &:nth-child(7n) {
+        margin-right: 0;
+      }
```
kuznetsova-anastasiia: does it make sense? as i see (devtools) this rule is always covered by another rules. mb i'm not right but it seems so

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+      margin-left: calcFirstDayMarginLeft($day-width-height, $day-num, $gap);
+    }
+
+    &--starting-day-#{$day-name} &__day:nth-child(7n) {
+      margin-right: $gap;
+    }
+
+    &--starting-day-#{$day-name} &__day:nth-child(7n + calc(7 - #{$day-num})) {
+      margin-right: 0;
+    }
```
kuznetsova-anastasiia: i see that you made setting gaps by margins in calendar difficult and hard to understand. maybe it would be simpler to use `gap: 1px` in `.calendar` since we don't have any tests. how do you think?

kuznetsova-anastasiia (approved):
you did a great job! please, check my advices below, maybe they will help you to make your code better :)

src/styles/blocks/_calendar.scss
stanislavcodes: true, unnecessary styles

src/styles/blocks/_calendar.scss
stanislavcodes: 👍You're right, it's an easier solution

stanislavcodes: > i see that you made setting gaps by margins in calendar difficult and hard to understand. maybe it would be simpler to use `gap: 1px` in `.calendar` since we don't have any tests. how do you think?

👍You're right, it's an easier solution

src/styles/blocks/_calendar.scss
stanislavcodes: > does it make sense? as i see (devtools) this rule is always covered by another rules. mb i'm not right but it seems so

true, unnecessary styles

src/styles/blocks/_calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+  padding: $calendar-paddings;
+
+  &__day {
+    @include square($day-width-height);
+    background-color: $color-day-bg;
+    border: 1px solid $color-day-border;
+    display: flex;
+    margin: 0 $gap $gap 0;
+    cursor: pointer;
+    font: $day-number-font-size "Arial", sans-serif;
```
lilia-obushenko: check the right order of css properties)
It will be more readable:)

lilia-obushenko (approved):
Good job! Just my advice, check css for order of properties)

Andr1yK (commented):
cool!

Andr1yK (approved):

PR #1718, opened 2023-01-29
solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/layout_calendar/)
[TEST REPORT LINK](https://ihor-zakharov.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,72 @@
+$square-size: 100px;
+$gap-size: 1px;
+$calendar-padding: 10px;
```
ViktorHavryliuk: Move all the variables to the separate file

src/styles/main.scss
```diff
@@ -1,3 +1,72 @@
...
+  &__day {
+    background-color: #eee;
+    width: $square-size;
+    height: $square-size;
+    border: 1px black solid;
+    display: flex;
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
```
ViktorHavryliuk: Move cursor: pointer to &__day itself

ViktorHavryliuk (changes requested):
Please fix modifiers naming (should be "--" instead of "_" between modifier and value, look checklist p.2)
image

PR #1808, opened 2023-02-14
create calendar

LBondarchuk:
- [DEMO LINK](https://LBondarchuk.github.io/layout_calendar/)
- [TEST REPORT LINK](https://LBondarchuk.github.io/layout_calendar/report/html_report/)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--31 calendar--start-day-mon">
```
agniya-i:
```suggestion
    <div class="calendar calendar--start-day--31 calendar--start-day--mon">
```
define modifier for the start day in this way

agniya-i (changes requested):
Please add all files to PR, I can't see the style files

src/utils/_extends.scss
```diff
@@ -0,0 +1,5 @@
+%cell-text-settings {
+  font-family: Arial, sans-serif;
+  font-size: 30px;
+  display: inline-block;
+}
```
Anastasiia-Svintsova, line 5: It's good that you know how to use it. But do you really need this extension here? You use it only once, you could write this style inside the body. 
You don't need to fix it now, just a comment for the next tasks so you don't do extra work when it's not needed

Anastasiia-Svintsova (changes requested):
Good job! Add some fixes 

- Align your calendar to the center, please 
image

- Why did you move the HTML file? It should be directly in the src folder
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+    &__cell {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      box-sizing: border-box;
+      width: 100px;
+      height: 100px;
+      background-color: #eee;
+      border: 1px solid black;
+      transition-duration: 0.5s;
```
polosanya, line 19: It is better to specify the transition-property too.
Now it applies to all the properties, which is not good for performance

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

Anastasiia-Svintsova (approved):

nazarmatsevych (approved):

witflash (changes requested):
Test

vlad-demchuk (approved):

sTorba24 (approved):
Good work!

PR #1854, opened 2023-02-28
added calendar

PavloReutskiy:
https://PavloReutskiy.github.io/layout_calendar/
https://PavloReutskiy.github.io/layout_calendar/report/html_report/


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+  &__month {
+    display: flex;
+    gap: $gap-between-days;
+    flex-wrap: wrap;
+    max-width: calendarWidth(
+      $day-size,
+      $gap-between-days,
+      $calendar-side-paddings
+    );
+    padding: 20px $calendar-side-paddings;
```
M1k1ta, line 14: Don't be greedy)). 20px is normal, but can be more. Also, you can use only padding-top or margin-top. I would prefer to use a margin-top.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,11 @@
+$day-size: 100px;
+$gap-between-days: 1px;
+$calendar-side-paddings: 10px;
```
M1k1ta, line 3: I think you create too a lot of variables. For example variables for 1px)) And I don't understand calendar-side-paddings. Why 10px, you can use numbers from the day of the week.

M1k1ta (approved):
Good, job!! If it works is already good.

src/styles/utils/_variables.scss
PavloReutskiy: - I decided to add all these variables for the function that limit the width of the container to protect it correct work
- I thought the same way about 1px)) but decided to add it in case somebody will decide to make gap bigger + I used it in  several places
- maybe I didn't understand task correct, but I take 10px here:
image

src/styles/blocks/calendar.scss
PavloReutskiy: generally agree! but here I decided to add padding 20px on the top and bottom and 10px - left and right; in case if I will use it again it will contain necessary paddings, but I'm not sure... maybe I'm wrong

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
+      <div class="calendar__month calendar__month--month-length--31">
+        <div class="calendar__day calendar__day--date--1 calendar__day--start-day--mon"></div>
```
SerhiiKirik: Try to do with fewer classes
Just use nth-child
```suggestion
      <div class="calendar__month calendar__month--month-length--31 calendar__day--start-day--mon">
        <div class="calendar__day"></div>
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+
+    @for $index from 28 through 30 {
+      &--month-length--#{$index} :nth-child(n+#{$index + 1}) {
+        display: none;
+      }
+    }
```
SerhiiKirik, line 20: move all your functions & mixins to separates files
```suggestion
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  &__day {
+    position: relative;
+    @include squere(100px);
+    border: 1px solid #000;
+    background-color: #eee;
+    transition: all 0.5s;
```
SerhiiKirik: don't use it like transition: all 0.3s
It's too hard for the browser
Check this
image

```suggestion
    transition: transform $transitionDuration;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+      &--month-length--#{$index} :nth-child(n+#{$index + 1}) {
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    position: relative;
+    @include squere(100px);
+    border: 1px solid #000;
```
SerhiiKirik: Create a variables for colors
```suggestion
    border: 1px solid $c-black;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+      transform: translateY(-20px);
+    }
+
+    @for $index from 1 through 31 {
+      &--date--#{$index} {
+        &::before {
+          content: "#{$index}";
+          @include positionCenter();
+          font-family: Arial, "Helvetica Neue", Helvetica, sans-serif;
+          font-size: 30px;
```
SerhiiKirik: move to html
```suggestion
```

SerhiiKirik (changes requested):
Great 🚀 
Let's improve your code & organize it 🧑‍💻

PavloReutskiy:
all done! thx, it looks much better!

SerhiiKirik (approved):
LGTM 🔥 
You make a great job, keep it up 🧑‍💻

PR #1860, opened 2023-02-28
selution

BondarenkoDmitriy:
- [DEMO LINK](https://BondarenkoDmitriy.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+    &--start-day--#{$day} :first-child {
+      margin-left: $dayNum * ($calendar-dey-box + $box-gap);
+    }
+  }
+
+  @for $i from 28 through 31 {
+    &--month-length--#{$i} :nth-child(n + #{$i + 1}) {
+      display: none;
+    }
+  }
```
VladVeklenko: Would be better if you create additional file for loops, for me it will be more accurate but I don't sure

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
+@import "../utils/_variables";
+
+@import "../utils/mixins";
+
+@import "../utils/normalize";
```
VladVeklenko: Also, you can use your main scss for import files, it will be more organized

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-box-size;
+  gap: $box-gap;
+  margin: 100px;
```
VladVeklenko: I don't exactly understand why you use margin of 100px. For me, in this case margin 0 auto will be more suitable

VladVeklenko (approved):
I think you have a good code in general, maybe mentor can say more about your mistake

src/styles/blocks/calendar.scss
VinnieJ-2k20: Yes, import these in main file. Also, no need for empty lines.
```suggestion
@import "../utils/_variables";
@import "../utils/mixins";
@import "../utils/normalize";
```

src/styles/blocks/calendar.scss
VinnieJ-2k20: No keep these to selectors as close as possible.

VinnieJ-2k20 (changes requested):
> Display a calendar in the middle of the screen.

Also, some more details in comments.

VinnieJ-2k20 (changes requested):
@BondarenkoDmitriy It looks like you're not reading comments.

1. Center the calendar on the page.
2. Move all imports to main style file - that's where all code is being composed.
3. Move loops with calendar styles to calendar file.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  &__day {
+    position: relative;
+    @include square($calendar-dey-box);
+
+    background-color: #eee;
+    border: 1px solid black;
+    transition-duration: 0.5s;
```
SerhiiKirik, line 42: don't use it like that, because it's the same as transition: all 0.5s
It's too hard for the browser
Check this
image

also, create a variable for transition Duration
```suggestion
    transition: transform $transitionDuration;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
...
+      display: none;
+    }
+  }
+
+  &__day {
+    position: relative;
+    @include square($calendar-dey-box);
+
+    background-color: #eee;
+    border: 1px solid black;
```
SerhiiKirik, line 41: don't forget to create variables for all colors
```suggestion
    border: 1px solid $c-black;
```

SerhiiKirik (approved):
Great 🔥 
Let's check the comments & move on
Also, your calendar is still not in a center :D
image

PR #1868, opened 2023-02-28
project calendar

Nazarii98:
[DEMO LINK](https://Nazarii98.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -3,7 +3,11 @@
   padding: 10px;
   display: flex;
   flex-wrap: wrap;
-  width: $widthCall;
+  gap: $calendarGapSize;
+  width: $calendarWidth;
+  color: black;
+  font-size: $fontSizeValue;
+  font-family: Arial, sans-serif;
 
   @each $dayName, $mgLeft in $days {
     &--start-day--#{$dayName} {
@@ -13,32 +17,27 @@
     }
   }
 
-  @for $index from 28 through 31 {
+  @for $index from 28 through $monthLength {
     &--month-length--#{$index} {
       $dayIndex: $index + 1;
       & > :nth-child(n + #{$dayIndex}) {
-        opacity: 0;
         display: none;
       }
     }
   }
 
   &__day {
-    background-color: $colorCell;
     box-sizing: border-box;
-    border: 1px solid black;
-    margin: 0.5px;
-    @include square(#{$daySize});
+    @include square(#{$cellSize});
+    background-color: $cellColor;
     text-align: center;
-    line-height: $daySize;
+    line-height: $cellSize;
+    border: 1px solid black;
     transition: color 0.5s, transform 0.5s;
 
     @for $index from 1 through $monthLength {
       &:nth-child(#{$index}):before {
         content: "#{$index}";
-        color: black;
-        font-size: $fontSizeNumber;
-        font-family: Arial, sans-serif;
       }
     }
 
@@ -46,7 +45,6 @@
       cursor: pointer;
       transform: translate(0, -$transformSize);
       background-color: pink;
-      animation-duration: scale(1.2);
     }
   }
 }
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,52 @@
+.calendar {
+  position: relative;
+  padding: 10px;
+  display: flex;
+  flex-wrap: wrap;
+  width: $widthCall;
+
+  @each $dayName, $mgLeft in $days {
+    &--start-day--#{$dayName} {
+      & > :first-child {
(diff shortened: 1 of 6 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+  }
+
+  @for $index from 28 through 31 {
+    &--month-length--#{$index} {
+      $dayIndex: $index + 1;
+      & > :nth-child(n + #{$dayIndex}) {
+        opacity: 0;
+        display: none;
```
roshirin: 'opacity: 0' does nothing here, 'display: none' would be enough

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  &__day {
+    background-color: $colorCell;
+    box-sizing: border-box;
+    border: 1px solid black;
+    margin: 0.5px;
+    @include square(#{$daySize});
+    text-align: center;
+    line-height: $daySize;
+    transition: color 0.5s, transform 0.5s;
```
roshirin: This and two following selectors don't match styles properties order (Checklist #12)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+        font-size: $fontSizeNumber;
+        font-family: Arial, sans-serif;
+      }
+    }
+
+    &:hover {
+      cursor: pointer;
+      transform: translate(0, -$transformSize);
+      background-color: pink;
+      animation-duration: scale(1.2);
```
roshirin: Looks like you forgot to remove the last line

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+    text-align: center;
+    line-height: $daySize;
+    transition: color 0.5s, transform 0.5s;
+
+    @for $index from 1 through $monthLength {
+      &:nth-child(#{$index}):before {
+        content: "#{$index}";
+        color: black;
+        font-size: $fontSizeNumber;
+        font-family: Arial, sans-serif;
```
roshirin: It would be better to assign these rules to an element rather than to all of its generated pseudo-classes. (Checklist #7)

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,18 @@
+$colorCell: #eee;
+
+$daySize: 100px;
+$countColomn: 7;
+$widthCall: $countColomn * ($daySize + 1px);
+$monthLength: 31;
+$lastDayToShow: 28;
+$fontSizeNumber: 0.3 * $daySize;
+$transformSize: 0.2 * $daySize;
```
roshirin: $lastDayToShow is not used anywhere.

$countColomn probably will stay constant while there are only 7 days in a week and it's used just once. It's an overuse of variables. The same is about $monthLength.

I think your variable names are somewhat confusing. Here are some of my suggestions:

- $colorCell and $daySize refer to the same object propereties - $cellColor and $cellSize would be better;
- $widthCall - $calendarWidth;
- $fontSizeNumber - $fontSizeValue.

(Checklist #6)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    background-color: $colorCell;
+    box-sizing: border-box;
+    border: 1px solid black;
+    margin: 0.5px;
```
roshirin: We can use 'gap: 1px' in this task because of no test check here.
If you want to use margins, 'margin-right: 1px' and 'margin-bottom: 1px' would be better.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,18 @@
...
+
+$days:
+  "mon" 0,
+  "tue" (($daySize + 1px) + 0.5px),
+  "wed" (($daySize + 1px) * 2 + 0.5px),
+  "thu" (($daySize + 1px) * 3 + 0.5px),
+  "fri" (($daySize + 1px) * 4 + 0.5px),
+  "sat" (($daySize + 1px) * 5 + 0.5px),
+  "sun" (($daySize + 1px) * 6 + 0.5px);
```
roshirin: Maybe it would be better to use a function for margin calculation. And this 0.5px intimidates alot.

roshirin (changes requested):
There are some issues

Nazarii98:
@roshirin Thanks for detailed checking my job and the good advice, it is very useful!

VinnieJ-2k20 (changes requested):
> Display a calendar in the middle of the screen.

One thing left to fix.

Nazarii98:
Done

VinnieJ-2k20 (approved):

PR #1897, opened 2023-03-02
calendar

KatiaGamal:
- [DEMO LINK](https://KatiaGamal.github.io/layout_calendar/)
- [TEST REPORT LINK](https://KatiaGamal.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,47 @@
+.calendar {
+  position: relative;
+  max-width: $cell-size * 8 + 6 * $cell-gap;
+  display: flex;
+  flex-wrap: wrap;
+  margin: 40px auto;
+  gap: $cell-gap;
+}
+
+.day {
+  height: $cell-size;
+  width: $cell-size;
+  border: 1px black solid;
+  background-color: #eee;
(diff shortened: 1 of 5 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  position: relative;
+  max-width: $cell-size * 8 + 6 * $cell-gap;
+  display: flex;
+  flex-wrap: wrap;
+  margin: 40px auto;
```
DarkMistyRoom, line 6: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  gap: $cell-gap;
+}
+
+.day {
```
DarkMistyRoom, line 10: make it a calendar's element

DarkMistyRoom (changes requested):
you know calendars? they should always have 7 columns. that's how they are
image

the modifiers don't work(
image

PR #1905, opened 2023-03-02
calendar

myplekan:
https://myplekan.github.io/layout_calendar/
https://myplekan.github.io/layout_calendar/report/html_report/


src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <ul class="calendar calendar--month-length--29 calendar--start-day--mon">
+        <li class="calendar__day calendar__day--1"></li>
```
vlad-demchuk: li should have only `calendar__day` class, you don't need `day--1, day--2` and so on

src/styles/utils/_loops.scss
```diff
@@ -0,0 +1,29 @@
+@for $dayIndex from 1 through 31 {
+  .calendar__day--#{$dayIndex}::before {
```
vlad-demchuk: you can use nth-child to avoid redundant classes

vlad-demchuk (changes requested):

nazarmatsevych (approved):
Well done 🔥

PR #1910, opened 2023-03-03
another attempt

AP-Solution:
[DEMO LINK](https://ap-solution.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,36 @@
+.calendar {
+  width: $calendar-width;
+  display: flex;
+  gap: $cell-gap;
+  flex-wrap: wrap;
+  &__cell {
+    transition: transform 0.5s;
+    cursor: pointer;
+    @include centeredSquare;
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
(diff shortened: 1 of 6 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,36 @@
...
+      }
+    }
+    border: $cell-border;
+    background-color: $cell-color;
+    &:hover {
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+}
```
SerhiiKirik, line 22: Don't forger about the empty line between classes
also, create variable for $transitionDuration
```suggestion
  flex-wrap: wrap;
  
  &__cell {
    transition: transform $transitionDuration;
    cursor: pointer;
    
    @include centeredSquare;
    
    @for $i from 1 through 31 {
      &:nth-child(#{$i})::before {
        content: "#{$i}";
      }
    }
    
    border: $cell-border;
    background-color: $cell-color;
    
    &:hover {
      background-color: pink;
      transform: translateY(-20px);
    }
  }
}
```

SerhiiKirik (approved):
Great 🔥 
Let's check the comments & move on
Also, you can align with center your calendar
image

PR #1914, opened 2023-03-04
calendar

Ilia991:
I do not understand what needs to be done to implement the modifier and why it is needed
https://Ilia991.github.io/layout_calendar/
https://Ilia991.github.io/layout_calendar/report/html_report/


etojeDenys (changes requested):
please, feel free to ask about it in the slack chat

PR #1944, opened 2023-03-11
SCSS practice

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,19 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta
+       name="viewport"
+       content="width=device-width, initial-scale=1.0"
+    >
     <title>Calendar</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link
+       rel="stylesheet"
+       href="./styles/main.scss"
+    >
   </head>
   <body>
     <div class="wrapper">
-      <div class="calendar calendar--start-day_mon calendar--month-length_31">
+      <div class="calendar calendar--start-day--mon calendar--month-length--31">
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
         <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
+      <div class="calendar calendar--start-day_mon calendar--month-length_31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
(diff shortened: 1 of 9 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+    transition: transform 0.5s, background-color 0.5s;
+    cursor: pointer;
+
+    &:hover {
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(2n + #{$i}) {
```
DorotaLeniecDev: 2n is redundant here, you can easly write:
```suggestion
      &:nth-child(#{$i}) {
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  @each $day, $number-of-day in $days {
+    &--start-day_#{$day} {
+      .calendar__day:first-child {
+        margin-left: get-margin-size($number-of-day);
+      }
+    }
+  }
```
DorotaLeniecDev: I think {}  are not needed here, you can remoe them like:
```suggestion
  @each $day, $number-of-day in $days {
    &--start-day_#{$day} .calendar__day:first-child {
        margin-left: get-margin-size($number-of-day);
      }
  }
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+
+  @for $i from 1 through 31 {
+    &--month-length_#{$i} {
+      .calendar__day:nth-child(n  + #{$i + 1}) {
+        display: none;
+      }
+    }
+  }
```
DorotaLeniecDev: Same here, {} are not needed before .calendar__day selector

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
+      <div class="calendar calendar--start-day_mon calendar--month-length_31">
```
DorotaLeniecDev: Modifier value should be added like this:
```suggestion
      <div class="calendar calendar--start-day--mon calendar--month-length--31">
```

src/styles/blocks/wrapper.scss
```diff
@@ -0,0 +1,7 @@
+.wrapper {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+  height: 100vh;
+  width: 100vw;
```
DorotaLeniecDev: Setting an element's width to 100vw makes it as wide as the current viewport. But doing that is bad practice. First, it is almost always unnecessary. A block element, by default, already takes up all the available width. Setting the width manually is not needed in that situation. Second, it may result in unnecessary horizontal scrolling as the page is wider than the visible view port width.

src/styles/blocks/wrapper.scss
```diff
@@ -0,0 +1,7 @@
+.wrapper {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+  height: 100vh;
```
DorotaLeniecDev, line 4: A good practice is also to add height: 100dvh, for dynamic height on mobiles (you can read more here https://css-tricks.com/the-large-small-and-dynamic-viewports/). Although dvh is not yet supported in all browsers: https://caniuse.com/?search=dvh but you can write it like 
```suggestion
  height: 100vh;
  height: 100dvh;
```

DorotaLeniecDev (changes requested):
Good job! Few minor changes needed

src/styles/blocks/calendar.scss
zibi-dev: Really weird. I don't know why i did that. It mus have been late hahaha

src/styles/blocks/wrapper.scss
zibi-dev: So if 100dvh isn't supported 100vh will be a fallback property?

DorotaLeniecDev (approved):
Well done!

src/styles/blocks/wrapper.scss
DorotaLeniecDev: yes exactly :)

src/styles/blocks/calendar.scss
DorotaLeniecDev: That is what Code Reviews are for

PR #1957, opened 2023-03-14
Develop

superzuc:
https://superzuc.github.io/layout_calendar/

https://superzuc.github.io/layout_calendar/report/html_report/


qoa11a (changes requested):
The user should be able to change starting day of the week. Currently, it isn't working.

PR #1975, opened 2023-03-16
calendar solution v1

bjeziorek:
- [DEMO LINK](https://bjeziorek.github.io/layout_calendar/)
- [TEST REPORT LINK](https://bjeziorek.github.io/layout_calendar/report/html_report/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,8 @@
+$weekdays: "mon" 1, "tue" 2, "wen" 3, "thu" 4, "fri" 5, "sat" 6, "sun" 7;
```
michalcywka: If we have a long variable, it is better to write it like so:
```suggestion
$weekdays: 
  "mon" 1, 
  "tue" 2,
  "wen" 3, 
  "thu" 4, 
  "fri" 5,
  "sat" 6,
  "sun" 7;
```

src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <section class="calendar calendar--monthLength-30">
```
michalcywka, line 10: Ideally we would like the whole calendar to control the starting day of the month. Also remember about BEM naming convention:
```suggestion
    <section class="calendar calendar__start-day--tue calendar__month-length--28">
```

src/index.html
```diff
@@ -6,7 +6,39 @@
...
+      <div class="calendar__day calendar__day--1 calendar__day--startDay-tue"></div>
```
michalcywka, line 11: Then here and in all of the divs below it we don't need information about which day it represents, it will simplify things a bit
```suggestion
      <div class="calendar__day day"></div>
```

src/styles/calendar.scss
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  &--monthLength-30 {
+    .calendar__day--31 {
+      display: none;
+    }
+    .calendar__day--31::before {
+      content: "";
+    }
+  }
```
michalcywka, line 36: This won't be necessary after some changes :)

michalcywka: What if we used a loop that generates these conditions for us:
```suggestion
    @for $i from 28 through 31 {
    &__month-length--#{$i} &__day:nth-child(#{$i}) ~ &__day{
      display: none;
    }
```

michalcywka (changes requested):
We can simplify things a bit :)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$weekdays:
+  "mon" 1,
+  "tue" 2,
+  "wen" 3,
+  "thu" 4,
+  "fri" 5,
+  "sat" 6,
+  "sun" 7;
```
michalcywka:
```suggestion
$weekdays:
  "mon" 0,
  "tue" 1,
  "wed" 2,
  "thu" 3,
  "fri" 4,
  "sat" 5,
  "sun" 6;
```

michalcywka: Also add `$gap: 1px`

src/styles/calendar.scss
michalcywka, line 1:
```
.day {
  width: $daySize;
  height: $daySize;
}

.calendar {
  width: $calendarWidth;
  display: flex;
  flex-wrap: wrap;
  gap: $gap;

  &__day {
    transition: background-color 0.3s, transform 0.3s;
    display: flex;
    align-items: center;
    justify-content: center;
    font-family: Arial, Helvetica, sans-serif;
    font-size: 30px;
    text-align: center;
    background-color: #eee;
    cursor: pointer;

    @for $day from 1 through 31 {
      &:nth-child(#{$day})::before {
        content: "#{$day}";
      }
    }

    &:hover {
      background-color: pink;
      z-index: 1;
      transform: translateY(-20px);
    }
  }

  @each $day, $number in $weekdays {
    &--startDay-#{$day} &__day:first-child {
      margin-left: $number * ($daySize + $gap);
    }
  }

  @for $i from 28 through 31 {
    &--monthLength-#{$i} &__day:nth-child(#{$i}) ~ &__day{
      display: none;
    }
  }
}

File: src/index.html (hunk 3 above)
Comment (michalcywka) [inline src/index.html]: All days should be the same
```suggestion
      <div class="calendar__day day"></div>
```

src/index.html
michalcywka:
```suggestion
   <section class="calendar calendar--monthLength-28 calendar--startDay-sun">
```

michalcywka (changes requested):
How to make it work :)

michalcywka (approved):

PR #1999, opened 2023-03-23
added task solution

mariuszgit:
https://mariuszgit.github.io/layout_calendar/
https://mariuszgit.github.io/layout_calendar/report/html_report/


src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
-body {
-  margin: 0;
-}
+@use "utils/reset" as *;
```
Radoslaw-Czerniawski: For the current version of the scss we are using, please use @import everywhere, example:
```
@use "utils/reset";
```

src/index.html
```diff
@@ -4,9 +4,43 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="stylesheet" href="styles.css">
```
Radoslaw-Czerniawski: Import here `./styles/main.scss` rather than `styles.css`

Radoslaw-Czerniawski (changes requested):
I tried to get your implementation working with @for 1 through 31 and changing the length in class. It didn't work most possibly because of you importing the style.css rather than main.scss. Correct the @use to @import, import main.scss instead of styles.css and adjust the pr to display 31 days as required in task requirements.

src/styles.css
DorotaLeniecDev, line 1: I think all .css files are unnecessary so they can be removed from PR

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="page">
+      <div class="calendar calendar--start-day_sun calendar--month-length_31">
```
DorotaLeniecDev: Follow BEM naming for modifier **block-name--modifier-name--modifier-value**
```suggestion
      <div class="calendar calendar--start-day--sun calendar--month-length--31">
```

DorotaLeniecDev (approved):
I have left 2 comments. One about removing css files from PR and other about BEM modifier class, it should look like this pattern: **block-name--modifier-name--modifier-value**. But everything looks good so I will approve it :)

PR #2025, opened 2023-04-02
add task solution

vovaAlieksieiev:
- [DEMO LINK](https://vovaAlieksieiev.github.io/layout_calendar/)


OleziO (changes requested):
image

Classname should contain a number that specifies the count of the month days.

OleziO (changes requested):
Changing month logic doesn't work
image

PR #2043, opened 2023-04-07
add task solution

kropiva:
https://kropiva.github.io/layout_calendar
https://kropiva.github.io/layout_calendar/report/html_report

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <div class="container__content calendar calendar--start-day--sun calendar--month-length--29">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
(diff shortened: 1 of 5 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
```
TarasHoliuk: It's better to use a variable for the gap (e.g., $gap: 1px) to make it consistent and easy to update in the future.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+  &__day {
+    width: $cell-size;
+    height: $cell-size;
+    background: #eee;
+    border: 1px solid black;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+    transition-duration: 0.5s;
```
TarasHoliuk: add `transition-properties` to list properties you really changing (it's better for performance than default `all` value)

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <div class="container__content calendar calendar--start-day--sun calendar--month-length--29">
```
```suggestion
      <div class="calendar container__calendar calendar--start-day--sun calendar--month-length--29">
```

src/styles/main.scss
```diff
@@ -1,3 +1,16 @@
...
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
+
+.container__content {
+  margin: 50px auto;
 }
```
TarasHoliuk: that's only 3 rows, but since it (container) is a separate BEM block it should be in a separate file
```suggestion
.container__calendar {
  margin: 50px auto;
}
```

TarasHoliuk (changes requested):
GJ :+1: 
Added a few minor comments

ivanchuk-yurii (approved):

PR #2045, opened 2023-04-08
Develop

zisuan-od:
https://zisuan-od.github.io/layout_calendar/


src/styles/main.scss
```diff
@@ -1,3 +1,11 @@
-body {
+@import "./utils/variables";
+@import "./utils/mixins";
+@import "./utils/functions";
+
+@import "./blocks/calendar";
+
+* {
   margin: 0;
```
VitaliyBondarenko1982, line 8: Avoid using global wildcard (*) selector for styling styles. Reset margin and padding only for elements where it is necessary
Only one property could be in * selector -` box-sizing: border-box`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,84 @@
...
+    justify-content: center;
+    align-items: center;
+    border: 1px solid grey;
+    border-radius: $cell-border-radius;
+    background-color: #eee;
+  }
+
+  &__day:hover {
+    cursor: pointer;
+    background-color: pink !important;
```
VitaliyBondarenko1982: Avoid using !important as it can make your CSS difficult to manage and maintain. Consider using a more specific selector or adjusting the order of your rules to increase specificity.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,84 @@
...
+  @for $index from 28 through 30 {
+    &__length--#{$index} {
+      .calendar__day:nth-child(n + #{1 + $index}) {
+        border: none;
+        background-color: lightgray;
+      }
+
+      .calendar__day:nth-child(n + #{1 + $index}):hover {
+        cursor: auto;
+        background-color: lightgray !important;
```
VitaliyBondarenko1982: Avoid using !important as it can make your CSS difficult to manage and maintain. Consider using a more specific selector or adjusting the order of your rules to increase specificity.

VitaliyBondarenko1982 (changes requested):
nice work!
please check a few comments

src/styles/main.scss
zisuan-od: I understand. But there are so many tasks, where it was asked to reset margins and paddings for all. Every mentor sees it in his own way. What is right?

zisuan-od: If i'll not reset margins and paddings, my calendar becomes some strange: 

image

zisuan-od: I'll leave margins and paddings in this task in this block

src/styles/blocks/calendar.scss
zisuan-od: it was so easy way to complete task faster... ok :)

src/styles/blocks/calendar.scss
zisuan-od: done

src/styles/main.scss
VitaliyBondarenko1982: I meant to reset margin and padding for specific tags, nor for all selectors.
in your case reset for `body` and `p`

VitaliyBondarenko1982 (approved):

PR #2058, opened 2023-04-12
add task solution

Dolnys:
[DEMO LINK](https://Dolnys.github.io/layout_calendar/)


anpiet (changes requested):
Line:  div class="calendar calendar--month-length-XX calendar--start-day-xyz">

According to BEM modifier value should be after '_' (not -). '-' is used in names of blocks, elements and modifiers.

Calendar should be in the middle of the screen, so it should be also aligned vertically.

Dolnys:
I've centered the calendar in both directions but I can't agree that the modifier value should be after '_'. According the materials: The name of modifier value should as per example:
block-or-element-name--modifier-name--modifier-value so if I should change something it is only to add "-" before the value, which is xx and xyz in my case. But there are probably many good ways you can do it, I don't assume that your method is wrong :)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  $days:
+    "mon" 0,
+    "tue" 1,
+    "wed" 2,
+    "thu" 3,
+    "fri" 4,
+    "sat" 5,
+    "sun" 6;
```
mvjl000: You could move this to variables file

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    cursor: pointer;
+    @include square($field-size);
+    background-color: $background-color;
+    box-sizing: border-box;
+    border: 1px solid $border-color;
+    font-size: 30px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    transition: all 0.5s;
```
mvjl000: Avoid setting transition to all, list all styles that you apply transition to. It's more readable and you know exactly what is affected by it.

mvjl000 (changes requested):
Minor changes needed. Besides of that I would like you to reorder your properties because it is a bit chaotic. You can read about it [here](https://github.com/necolas/idiomatic-css#declaration-order)

src/styles/blocks/calendar.scss
Dolnys: Done

src/styles/blocks/calendar.scss
Dolnys: Replace 'all' with 'transform' and 'background color

Dolnys:
Done. I hope it's okay now.👍

choeqq (approved):
All good 🚀 Nicely done!

PR #2076, opened 2023-04-17
add task solution

maksym-nemera:
- [DEMO LINK](https://maksym-nemera.github.io/layout_calendar/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,2 @@
+$width-and-height-for-box: 100px;
+$gap-for-box: 1px;
```
naziamoff: I'd say that `size` and `gap` will be good enough here, so it's overkilling

naziamoff (changes requested):
Great job actually) But there's one thing to fix:

I believe there is exactly 7 days in the week, not depending on screen scale 😄 
Screenshot 2023-04-20 at 17 19 23

naziamoff (approved):
nailed it 💅

PR #2078, opened 2023-04-18
Calendar

anpiet:
[DEMO LINK](https://anpiet.github.io/layout_calendar/)


Dolnys (approved):
Everything looks good except for the hover effect. It needs a small adjustment to ensure a smooth transition back to its initial state. Other than that, everything is fine.

src/index.html
```diff
@@ -7,6 +7,41 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <main class="main-page">
+      <div class="calendar calendar--start-day_sun calendar--month-length_29">
```
mvjl000, line 11: From the checklist:

- BEM - Make sure to follow BEM naming convention for complex modifiers: `block-name--modifier-name--modifier-value`

Let's remove that underscore and use a dash ;)

src/styles/main.scss
mvjl000, line 1: I need you to make some changes to styles structure

- You should separate styles to different files. Create new directory inside styles folder i.e. blocks in which you will create file for calendar block `calendar.scss`. All of your styles that are calendar-related should be there.
- Inside styles create new file for variables. It will be a lot cleaner!
- Then, import everything to `main.scss` file (btw you should put there also styles for body 😉)

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+    height: $Day-box-height;
+    border: $Day-box-border solid $Main-color;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background-color: #eee;
+    position: relative;
+
+    &:hover {
+      // bottom: 20px;
```
mvjl000: Remove unnecessary comments

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+      transform: translateY(-20px);
+      background-color: pink;
+      transition-duration: 500ms;
+      cursor: pointer;
```
mvjl000: I would place `cursor: pointer;` inside of `&__day-box`, not hover.
Same thing with transition, it should be there because right now, the animation is broken. Move it up and you will se nice and smooth animation.

And also make sure to list in transition all styles that you apply transition to. It's more readable and you know exactly what is affected by it.

So, in hover you should have only transfromn and background-color ;)

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  &__day-box {
+    box-sizing: border-box;
+    width: $Day-box-width;
+    height: $Day-box-height;
+    border: $Day-box-border solid $Main-color;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background-color: #eee;
+    position: relative;
```
mvjl000: You don't need position relative in here. You are using flex to position before and transform to animate on hover

src/index.html
```diff
@@ -7,6 +7,41 @@
...
+        <div class="calendar__day-box calendar__day-box--22"></div>
+        <div class="calendar__day-box calendar__day-box--23"></div>
+        <div class="calendar__day-box calendar__day-box--24"></div>
+        <div class="calendar__day-box calendar__day-box--25"></div>
+        <div class="calendar__day-box calendar__day-box--26"></div>
+        <div class="calendar__day-box calendar__day-box--27"></div>
+        <div class="calendar__day-box calendar__day-box--28"></div>
+        <div class="calendar__day-box calendar__day-box--29"></div>
+        <div class="calendar__day-box calendar__day-box--30"></div>
+        <div class="calendar__day-box calendar__day-box--31"></div>
```
mvjl000, line 42: Those 'boxes' should have only one class, in this case `calendar__day-box`. You shouldn't be indicating the order of boxes by a modificator. In styles you can make it work with help of properly implemented loops

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.calendar {
+  margin: auto;
```
mvjl000: This margin is redundant. Calendar is already centered in body because of flex used there.

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: 1px;
+  width: 706px;
```
mvjl000: This 706 is like a magic number, it is not clear why is it set to 706px and if you change for instance dimensions of calendar-day things will break up.

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  padding: 10px;
+
+  $days:
+    "tue" 101px,
+    "wed" 202px,
+    "thu" 303px,
+    "fri" 404px,
+    "sat" 505px,
+    "sun" 606px;
```
mvjl000: Don't set such a values. Instead of pixels just put the order of days:
```suggestion
  $days:
    "mon" 1,
    "tue" 2,
    "wed" 3,
// etc...
```

mvjl000 (changes requested):
You need to make some logic changes here. And one more thing I didn't mention in comments is [the order of properties](https://github.com/necolas/idiomatic-css#declaration-order) - check it out and apply it to your styles ;)

src/index.html
anpiet: On video lessons we were learned to use values of modifiers with underscore:
block-name__element-name--modifier-name_modifier-value
so in the case **calendar--start-day_sun** and **calendar--month-length_29** are OK.
Please check BEM video: https://prnt.sc/fY_2AOWUhfLb

src/index.html
anpiet: I have used this not for order, but to put a day number into the box with '::before'. I can change this and put day numbers directly in the html.

mvjl000 (approved):
🚀

PR #2089, opened 2023-04-23
Solution

yao157:
[DEMO LINK](https://yao157.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  @for $i from 28 to 31 {
+    &--month-lenght--#{$i} :nth-child(n+#{$i+1}) {
+      display: none;
+    }
+  }
+  @for $day from 1 through 31 {
+    &__day:nth-child(#{$day})::before {
+      box-sizing: border-box;
+      content: "#{$day}";
```
vlad-demchuk: image

vlad-demchuk: leave only `content: "#{$day}";` in loop

no need to repeat all styles

vlad-demchuk (changes requested):

src/index.html
```diff
@@ -4,9 +4,43 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="stylesheet" href="./styles/blocks/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
```
DarkMistyRoom: Would you happen to really need this container? you can just set all flex properties to the calendar directly and it will work the same way, with fewer tags

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,13 @@
...
+$base-color: #eee;
+
+$days:
+  "mon" 0,
+  "tue" $sell-size + $gap,
+  "wed" 2 * $sell-size + 2 * $gap,
+  "thu" 3 * $sell-size + 3 * $gap,
+  "fri" 4 * $sell-size + 4 * $gap,
+  "sat" 5 * $sell-size + 5 * $gap,
+  "sun" 6 * $sell-size + 6 * $gap;
```
DarkMistyRoom: to avoid repeating this formula for each day set it once in the loop where you apply this variable

DarkMistyRoom (changes requested):
Good job, just a few suggestions:

vlad-demchuk (changes requested):
start day modifier doesn't work now

vlad-demchuk (approved):

PR #2090, opened 2023-04-24
solution

serentinos:
- [DEMO LINK](https://serentinos.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,46 +1,44 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <div class="calendar calendar--start-day--mon calendar--month-length--30">
-      <ul class="calendar__month">
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
-        <li class="calendar__day"></li>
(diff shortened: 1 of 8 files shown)


src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--mon calendar--month-length--30">
+      <div class="calendar__month">
```
andrusb21: i think it will be better to use semantic tags

andrusb21 (approved):
GJ

src/index.html
serentinos: Thanks! I forgot about that.

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--mon calendar--month-length--30">
+      <ul class="calendar__month">
```
Anna-Tkachenko: Do you need extra wrapper here? Let's get rid of it, you've already have `div` as wrapper for the calendar

Anna-Tkachenko (changes requested):

niaYaro (approved):
Great job! 🤟 

According to the task 'The hovered cell has to become pink', but you used another color. It's not critical. Just for your attention.

serentinos:
> Great job! love_you_gesture
> 
> According to the task 'The hovered cell has to become pink', but you used another color. It's not critical. Just for your attention.

@niaYaro Just wanted to make it a little more unique :)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$field-width: ($day-cell-size + $cells-gap) * 7;
+
+$start-day: (
+  "mon": 0,
+  "tue": ($day-cell-size + $cells-gap),
+  "wed": ($day-cell-size + $cells-gap) * 2,
+  "thu": ($day-cell-size + $cells-gap) * 3,
+  "fri": ($day-cell-size + $cells-gap) * 4,
+  "sat": ($day-cell-size + $cells-gap) * 5,
+  "sun": ($day-cell-size + $cells-gap) * 6
```
Anna-Tkachenko, line 14: you can move calculation in the cycle where you set the correct properties for each of this modifiers

Anna-Tkachenko (approved):
Good job!

PR #2099, opened 2023-04-25
add task solution

niaYaro:
- [DEMO LINK](https://niaYaro.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="content">
+    <section class="content__calendar calendar calendar--start-day-mon calendar--month-length-31">
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
+      <div class="calendar__date"></div>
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="content">
+    <section class="content__calendar calendar calendar--start-day-mon calendar--month-length-31">
```
Yuliya1907, line 10: You don't need to make content as a BEM-block.It used only for positioning the calendar on the page.

Yuliya1907 (approved):
Good Job!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,70 @@
...
+      };
+    }
+  }
+
+  $weekdays:
+    "mon" 0,
+    "tue"
+    calendarSize($squareSize, $calendarMainPadding, $calendarGap)
+    - 6 * $squareSize - 2 * $calendarMainPadding - 5 * $calendarGap,
+    "wed"
```
Anna-Tkachenko: move this calculation in the cycle where you set needed properties for each modifier
For now this code looks too complicated

Anna-Tkachenko (changes requested):

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+        font-family: $calendarFont;
+        font-size: 30px;
+      };
+    }
+  }
+
+  $weekdays:
```
Anna-Tkachenko, line 38: Will be better to move it to variables file

Anna-Tkachenko (approved):
Good job! ⭐️

PR #2106, opened 2023-04-26
add task calendar  solution

booobryyyk:
[DEMO LINK](https://maksym-bobryk.github.io/layout_calendar/)
[TEST REPORT LINK](https://maksym-bobryk.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,14 @@
+@import "./utils/variables.scss";
+@import "./utils/mixins.scss";
+@import "./blocks/calendar.scss";
+
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
   margin: 0;
+
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
 }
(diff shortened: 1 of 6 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,4 @@
+$block-size: 100px;
+$block-gap: 1px;
+$grey-color: #eee;
```
amashutikov, line 3: We are using this color only once. So i think there is no need to make separete variable for it.

amashutikov (approved):
Good job! Everything looks nice to me)

src/styles/main.scss
```diff
@@ -1,3 +1,14 @@
+@import "./utils/variables.scss";
+@import "./utils/mixins.scss";
+@import "./blocks/calendar.scss";
+
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
```
Anna-Tkachenko, line 9: Will be better to move it to some class, instead of setting all styles directly to `body` tag

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #2111, opened 2023-04-26
add task solution calendar

annholovko:
[DEMO LINK](https://annholovko.github.io/layout_calendar/)

### DIFF
--- a/src/blocks/content.scss
+++ b/src/blocks/content.scss
@@ -1,0 +1,7 @@
+.content {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  padding: 30px;
+}
--- a/src/utils/_functions.scss
+++ b/src/utils/_functions.scss
@@ -1,0 +1,3 @@
+@function container-size($cell-size, $cell-gap, $container-padding) {
+  @return 7 * $cell-size + 6 * $cell-gap + 2 * $container-padding;
+}
--- a/src/utils/_variables.scss
+++ b/src/utils/_variables.scss
@@ -1,0 +1,16 @@
+$cell-size: 100px;
+$container-padding: 10px;
+$cell-gap:1px;
+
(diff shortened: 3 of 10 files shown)


src/utils/_functions.scss
```diff
@@ -0,0 +1,3 @@
+@function container-size($cell-size, $cell-gap, $container-padding) {
+  @return 7 * $cell-size + 6 * $cell-gap + 2 * $container-padding;
+}
```
ulositska, line 3: The best solution for always having 7 days in a week, especially the function taking in multiple parameters,it gives your calendar a great deal of flexibility 🤩

src/utils/_variables.scss
```diff
@@ -0,0 +1,16 @@
...
+$hover-color: #ffc0cb;
+
+$weekdays:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
```
ulositska, line 16: For some reason, I haven't thought of moving $weekdays variable into variables file. But it makes perfect sense!!

src/blocks/content.scss
```diff
@@ -0,0 +1,7 @@
+.content {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  padding: 30px;
+}
```
ulositska, line 7: You can move it to the variables file as well. It's used only once, but it may get difficult to find it, if you want to change it in the future, what do you think?

ulositska (approved):
The code is well-ordered, clean, all modifiers work as expected, the calendar doesn't shrink if the screen gets smaller 🔥🔥🔥🔥 

Quit honestly, you left me no job 😿

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #2112, opened 2023-04-26
add task solution

MaksimKad:
- [DEMO LINK](https://MaksimKad.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+      transform: translate(0, -20px);
+    }
+
+    @for $i from 1 through 31 {
+      &--#{$i}::before {
+        content: "#{$i}";
+
+        font-family: Arial, Helvetica, sans-serif;
+        font-size: 30px;
```
skinnyjazz: You should not prescribe the same styles 31 times. Take it out of the loop.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  }
+
+  @each $day, $day-num in $days {
+    &__start-day--#{$day} &__day--1 {
+      margin-left: ($days-cell-size * ($day-num - 1)) + ($gap * ($day-num - 1));
```
skinnyjazz: This can be written more compactly
(days-cell-size + gap) * day-num  
also, you can start day num list from 0, not from 1.

src/styles/functions.scss
```diff
@@ -0,0 +1,4 @@
+@function size($width, $gap) {
+  @return ($width * 8) + ($gap * 7);
```
skinnyjazz: Why 8? 
You have a lot of extra space because you didn't count 20px of padding

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+
+    list-style: none;
+
+    width: $days-cell-size;
+    height: $days-cell-size;
+    background-color: #eee;
```
skinnyjazz: You have variable $cell-color, but you don't use it

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0;
+  padding: 10px;
```
skinnyjazz: Padding should be stored in a variable in order to access it when calculating the width.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  gap: $gap;
+  width: size($days-cell-size, $gap);
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+
+    list-style: none;
```
ivankhrsh, line 18: you can move it to normalize file

src/styles/blocks/calendar.scss
ivankhrsh: [Check list №11](https://github.com/mate-academy/layout_calendar/blob/master/checklist.md) [STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.

So it's means that padding should be equal or bigger than value that you use to move calendar days. 
` transform: translate(0, -20px);`
Also you can move this value to variables.

src/styles/functions.scss
ivankhrsh: Don't got what this numbers 8 and 7 means and why we use it, if it's something important it's better to make variables with understandable names.

ivankhrsh (changes requested):

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <ul class="calendar calendar__start-day--fri calendar__month-length--30">
+      <li class="calendar__day calendar__day--1"></li>
```
Anna-Tkachenko: You don't need to specify separate class for each `li` element with `li` number
Use `nth-child` instead of separate class

Anna-Tkachenko (changes requested):

src/styles/normalize.css
```diff
@@ -0,0 +1,7 @@
+* {
```
olehmahera, line 1: we should never use *, just count your tags (ul, li)

olehmahera (changes requested):
Display a calendar in the middle of the screen.
Everything else looking good.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0;
```
Ihor-Karpyn: When using BEM, it is recommended not to include external styles such as margins or positions within the BEM block.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+  padding: $calendar-padding;
+  padding-top: $calendar-top-padding;
```
Ihor-Karpyn: Combine to one rule

src/styles/blocks/calendar.scss
Ihor-Karpyn: agree

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    cursor: pointer;
+
+    list-style: none;
+
+    width: $days-cell-size;
+    height: $days-cell-size;
```
Ihor-Karpyn: Mixin square will be useful here

src/styles/functions.scss
```diff
@@ -0,0 +1,4 @@
+@function size($cell-width, $cell-gap, $container-padding) {
```
Ihor-Karpyn: Size of what?

Ihor-Karpyn (changes requested):
GJ, almost done

src/styles/blocks/calendar.scss
Ihor-Karpyn: image

src/styles/mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin cell-size {
+  width: $days-cell-size;
+  height: $days-cell-size;
+}
```
Ihor-Karpyn, line 4:
```suggestion
@mixin square($size) {
  width: $size;
  height: $size;
}
```

Ihor-Karpyn (approved):
GJ, left few suggestions

PR #2121, opened 2023-04-27
Develop

ivankhrsh:
- [DEMO LINK](https://ivankhrsh.github.io/layout_calendar/)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sat calendar--month-length-30">
+      <div class="calendar__cell calendar__cell--1"></div>
```
Anna-Tkachenko: Don't set separate class for each day
Use `nth-child` instead of setting separate classes

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! ⭐️

PR #2122, opened 2023-04-27
Develop

YeMyr:
- [DEMO LINK](https://yemyr.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,10 +1,3 @@
-.container {
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  height: 100%;
-}
-
 .calendar {
   display: flex;
   flex-wrap: wrap;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -20,6 +20,8 @@
     border: 1px solid #000;
     background-color: #eee;
     cursor: pointer;
+    transition-duration: 2s, 500ms;
+    transition-property: transform, background-color;
 
     &:hover {
       background-color: pink;
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,53 @@
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+}
+
+.calendar {
+  display: flex;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    min-width: $day-width;
+    min-height: $day-height;
+    border: 1px solid #000;
+    background-color: #eee;
+    cursor: pointer;
+
+    &:hover {
+      background-color: pink;
+      transform: translateY(-20px);
+      transition: all 0.5s;
```
dimoshen, line 22: check (p. 9) [https://github.com/mate-academy/layout_calendar/blob/master/checklist.md](url) from the readme, and don't let your blocks drop like bricks:)

dimoshen (changes requested):
GJ Mate) I have only one remark about your hover

src/styles/blocks/calendar.scss
YeMyr: Yep, my bad))) thanks!
thanks!)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,55 @@
+.container {
```
silchyn: this is another block

src/styles/blocks/calendar.scss
silchyn: not fixed yet

silchyn (changes requested):

src/styles/blocks/calendar.scss
YeMyr: @silchyn 
image
Why?

src/styles/blocks/calendar.scss
YeMyr: fixed
image

src/styles/blocks/calendar.scss
silchyn: I also don't know why you keep this `transition` property for `:hover` when you already have it set

silchyn (approved):

src/styles/blocks/calendar.scss
YeMyr: It's just a different transition time when the cursor is hovered over the element and when it is removed from it.

PR #2161, opened 2023-05-02
add task layout_calendar

msukhovych:
demo  https://msukhovych.github.io/layout_calendar/
test  https://msukhovych.github.io/layout_calendar/report/html_report/


qoa11a (changes requested):
image

Fix the layout for all days other than Sunday.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,31 @@
...
+    justify-content: center;
+    align-items: center;
+    flex-direction: column;
+    margin-block-end: $margin-day;
+    margin-inline-end: $margin-day;
+    width: $size-day;
+    height: $size-day;
+    border: 1px solid #000;
+    background-color: #eee;
+    transition-duration: 0.5s;
```
polosanya, line 23: It would be great to add transition properties too. Now it applies to all properties, which is not good for performance
Also, let's make transition duration a variable. Same with all colors

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  &__day:hover {
+    cursor: pointer;
```
polosanya, line 27: You can add this property to the default state of the element

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #2177, opened 2023-05-09
add task solution

FIREDAN99:
https://FIREDAN99.github.io/layout_calendar/
https://FIREDAN99.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -5,6 +5,7 @@
   width: $col-count * $cell-size + $calendar-padding;
 
   &__day {
+    display: block;
     @include square($cell-size);
     box-sizing: border-box;
     border: 1px solid #000;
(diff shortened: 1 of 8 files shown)


mykhalenych (changes requested):
Screenshot 2023-05-09 at 18 35 34
you have created a redundant element

etojeDenys (changes requested):
i should have the opportunity to change the number of cells just by changing the modifier

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,60 @@
...
+        display: none;
+      }
+    }
+
+    @if ($days-count == 30) {
+      .calendar__day--31 {
+        display: none;
+      }
+    }
+  }
```
anstsot: please change the CSS a bit because it's not working

anstsot (changes requested):
Good work, but one of the conditions of the task was not met.
Please fix the modifier `month-length`

mykhalenych (approved):

PR #2205, opened 2023-05-18
Add calendar task solution

Serhii-Dr:
https://Serhii-Dr.github.io/layout_calendar/
https://Serhii-Dr.github.io/layout_calendar/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,12 +1,47 @@
 <!DOCTYPE html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--mon calendar--month-length--31">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,12 +1,47 @@
...
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+    </div>
   </body>
-</html>
+
+</html>
```
polosanya, line 47:
```suggestion
</html>

```
Don't forget to add newlines at the end
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

src/styles/_calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+  @for $day-length from 28 through 31 {
+    &--month-length--#{$day-length} &__day:nth-child(n + #{$day-length + 1}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    @include flex;
+    @include square($width, $height);
+    transition: all 0.5s;
```
polosanya, line 28: It is better to specify the exact properties instead of 'all'. It will be better for the performance
Also, it would be great to create a variable for transition duration

polosanya (approved):
Good job!

IvanFesenko (approved):

PR #2222, opened 2023-05-23
finished task

artkas03:
- [DEMO LINK](https://artkas03.github.io/layout_calendar/)


And678 (changes requested):
Good work, but one thing to consider, in requirements it is written : `DON'T add numbers in HTML (you will do it using CSS)`. Instead of specifying classes like `.calendar__day--25`, just use `:nth-child` in css.

tetiana-ivanchenko (approved):
GJ 🔥

PR #2225, opened 2023-05-24
Calendar solution

ArtemVelychko:
- [DEMO LINK](https://ArtemVelychko.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,56 @@
+.calendar-wrapper {
+  display: flex;
+  padding: 100px 100px;
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+  width: calendarWidth($calendarWidth, $calendarPadding);
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendarGap;
+
+  @each $day, $startDay in $days {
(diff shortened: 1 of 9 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+      display: none;
+    }
+  }
+
+  &__day {
+    @include daySize($calendarWidth);
+    background-color: #eee;
+    border: 1px solid black;
+    box-sizing: border-box;
+    transition: all 0.5s;
```
polosanya: It's not recommended to set 'all' as transition property. Better specify the exact properties 
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    &::before {
+      display: flex;
+      height: 100%;
+      justify-content: center;
+      align-items: center;
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      content: "";
```
polosanya: Pay attention to CSS properties order
https://9elements.com/css-rule-order/

polosanya (changes requested):
Almost done! Let's try to center the calendar horizontally
image

qoa11a (approved):

PR #2229, opened 2023-05-24
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/layout_calendar/)


And678 (changes requested):
Good work, but one thing to consider, in requirements it is written : `DON'T add numbers in HTML (you will do it using CSS)`. Instead of specifying classes like `.calendar__day--25`, just use `:nth-child` in css.

ollavka:
Hi, thank you)
I thought that by "DON'T add numbers in HTML (you will do it using CSS)" they mean not to add numbers in html right inside the tag, and the number modifier will not be considered as an error😁

tetiana-ivanchenko (approved):
GJ 🔥

PR #2230, opened 2023-05-24
part of the task is done

h1deeeee:
- [DEMO LINK](https://h1deeeee.github.io/layout_calendar/)
I did part of the task and then I got stuck and I can’t figure out what to do next


volodymyr-soltys97 (changes requested):
Good start 👍 
But let's first center the calendar in the middle of the page
image
You need to write a modifier through classes, with which we can change the number of days in the month from 28 to 31 and you can also change the beginning of the month to any day of the week from Monday to Sunday

PR #2235, opened 2023-05-25
add calendar solution

Medaed:
- [DEMO LINK](https://Medaed.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Medaed.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,38 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar calendar--start-day--sun calendar--month-length--29">
-      <div class="calendar__block calendar__block--1"></div>
-      <div class="calendar__block calendar__block--2"></div>
-      <div class="calendar__block calendar__block--3"></div>
-      <div class="calendar__block calendar__block--4"></div>
-      <div class="calendar__block calendar__block--5"></div>
-      <div class="calendar__block calendar__block--6"></div>
-      <div class="calendar__block calendar__block--7"></div>
-      <div class="calendar__block calendar__block--8"></div>
-      <div class="calendar__block calendar__block--9"></div>
-      <div class="calendar__block calendar__block--10"></div>
-      <div class="calendar__block calendar__block--11"></div>
-      <div class="calendar__block calendar__block--12"></div>
-      <div class="calendar__block calendar__block--13"></div>
-      <div class="calendar__block calendar__block--14"></div>
-      <div class="calendar__block calendar__block--15"></div>
-      <div class="calendar__block calendar__block--16"></div>
-      <div class="calendar__block calendar__block--17"></div>
-      <div class="calendar__block calendar__block--18"></div>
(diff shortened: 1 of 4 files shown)


polosanya (changes requested):
Please don't forget to add demo link to the PR description
image

src/index.html
```diff
@@ -7,6 +7,38 @@
...
+      <div class="calendar__block calendar__block--22"></div>
+      <div class="calendar__block calendar__block--23"></div>
+      <div class="calendar__block calendar__block--24"></div>
+      <div class="calendar__block calendar__block--25"></div>
+      <div class="calendar__block calendar__block--26"></div>
+      <div class="calendar__block calendar__block--27"></div>
+      <div class="calendar__block calendar__block--28"></div>
+      <div class="calendar__block calendar__block--29"></div>
+      <div class="calendar__block calendar__block--30"></div>
+      <div class="calendar__block calendar__block--31"></div>
```
DarkMistyRoom: No need to create a specific modifier for each day. Leave there only the `calendar__block` class and use nth-child to work with them

src/styles/main.scss
```diff
@@ -1,3 +1,57 @@
...
+
 body {
   margin: 0;
 }
+
+.calendar {
+  height: 100vh;
+  width: 50%;
+  display: flex;
+  margin: 0 auto;
```
DarkMistyRoom: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/main.scss
```diff
@@ -1,3 +1,57 @@
...
+  gap: 1px;
+
+  &__block {
+    box-sizing: border-box;
+    height: 100px;
+    width: 100px;
+    border-color: #eee;
+    border: 1px solid black;
+    padding: 10px;
+    transition: all 0.5ms;
```
DarkMistyRoom: Transition doesn't work. And list all transition properties, it is better for performance

DarkMistyRoom (changes requested):
- What about the cells' background color? It should be grey

image
image

- Your calendar has 8 days) What would we call the new day of the week?)
- But now it has only 2 days. Please fix it. There always should be 7 days
image

- Add all repeated values to variables

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+        line-height: 90px;
+      }
+    }
+  }
+
+  &__block:hover {
+    cursor: pointer;
+    background-color: pink;
+    transform: translateY(-20px);
+  }
```
VitaliyBodnar: the hover should be inside the &__block
please check the checklist:
'[SASS] - Make use of SASS nesting - write pseudo-class, pseudo-element selectors inside general selector. As well as media queries.'

VitaliyBodnar (changes requested):
Good Job 🔥, let's fix a little issue.

SanyaBratashchuk (approved):

PR #2243, opened 2023-05-27
add task solution

tarovyk:
- [DEMO LINK](https://purpleFade.github.io/layout_calendar/)
- [TEST REPORT LINK](https://purpleFade.github.io/layout_calendar/report/html_report/)


SanyaBratashchuk (changes requested):
I should be able to set the first day of the month via modifier

PR #2247, opened 2023-05-29
Solution layout_calendar

Svitlanchick:
https://github.com/mate-academy/layout_calendar/pull/2247


src/index.html
```diff
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <h1></h1>
```
witflash: If you have removed the text don't leave an empty tags. Or just remove the tag or keep the text.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    text-align: center;
+    height: $daySize;
+    width: $daySize;
+    font-family: $BodyFont;
+    size: $daySize;
+    gap: $dayGap;
```
witflash: Do these props really work?
You have to use `width`, `height`, `margin-right` and `margin-botton` properties instead of these.
For example:
```
width: $daySize;
height: $daySize;
margin-right: $dayGap;
margin-bottom: $dayGap;
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+    background-color: #eee;
+    border: 1px solid black;
+    color: black;
+    transition: all 0.5s ease-in-out;
```
witflash: Don't use `transition` for `all`. Use needed properties only:
```
transition: transform 0.5s ease-in-out;
```

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
+// $BodyFont: arial, helvetica, sans-serif;
+// $FontSize: 30px;
+// $daySize: 100px;
+// $dayGap: 1px;
+// $totalSize: 7 * $daySize;
+// $totalGap: 6 * $dayGap;
```
witflash: Remove all commented code

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+@import "../styles/utils/variables";
+@import "./blocks/calendar.scss";
+
 body {
+  box-sizing: border-box;
   margin: 0;
 }
+
+// .calendar {
```
witflash: The same here. Remove all comments.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$BodyFont: arial, helvetica, sans-serif;
+$FontSize: 30px;
+$daySize: 100px;
```
witflash: Use the same naming for every variable. It's better to use `camelCase` everywhere. Don't use different approaches.

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
+// $BodyFont: arial, helvetica, sans-serif;
+// $FontSize: 30px;
+// $daySize: 100px;
+// $dayGap: 1px;
+// $totalSize: 7 * $daySize;
+// $totalGap: 6 * $dayGap;
+@import "../styles/utils/variables";
```
witflash: There is no need to use such an import.
Use the next way:
```
@import "./utils/variables";
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  width: $totalSize + $totalGap + 10px;
+  padding: 10px;
+
+  $days: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
+
+  @each $day, $number in $days {
+    &--start-day--#{$day} .calendar__day:first-child {
+      margin-left: ($number * ($daySize + $dayGap)) + 3 * $dayGap;
```
witflash: You don't need use `3 * $dayGap` at the end. If you need this there is some mistake with your width calculation. Check it, please

witflash (changes requested):
Check my comments, please

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$BodyFont: arial, helvetica, sans-serif;
+$FontSize: 30px;
+$DaySize: 100px;
+$DayGap: 1px;
+$TotalSize: (7 * $DaySize);
+$TotalGap: 7 * $DayGap;
```
witflash, line 6: It's better to name all variables with small letters at the start. It's a common rule for `camelCase`.

witflash (approved):
Well done 👍

PR #2264, opened 2023-06-01
add task solution

MariiaMonakhova:
- [DEMO LINK](https://MariiaMonakhova.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,9 +1,9 @@
 .calendar {
-  width: 7.1 * $box-width;
-  height: 5.1 * $box-height;
+  width: 7 * $box-width + 6px;
 
   display: flex;
   flex-wrap: wrap;
+  justify-content: flex-start;
   gap: 1px;
 
   padding: 10px;
@@ -29,7 +29,6 @@
   }
 
   &__box {
-    position: relative;
     box-sizing: border-box;
 
     height: $box-height;
@@ -38,24 +37,18 @@
     background-color: #eee;
     border: 1px solid black;
     text-align: center;
+    cursor: pointer;
+    line-height: 100px;
 
     transition-duration: 0.5s;
 
     @for $box from 1 through $box-col {
       &:nth-child(#{$box})::before {
         content: "#{$box}";
-        position: absolute;
-        left: 50%;
-        top: 50%;
-        transform: translate(-50%, -50%);
-
-        font-size: 30px;
-        font-family: Arial, Helvetica, sans-serif;
       }
     }
 
     &:hover {
-      cursor: pointer;
       background-color: pink;
 
       transform: translateY(-20px);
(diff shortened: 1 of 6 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
+.calendar {
+  width: 7.1 * $box-width;
+  height: 5.1 * $box-height;
```
anstsot: why are you using float values like `7.1`? maybe it would be better to multiply on 7 and add exact `padding`/`margin`/`gap`?

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+    @for $box from 1 through $box-col {
+      &:nth-child(#{$box})::before {
+        content: "#{$box}";
+        position: absolute;
+        left: 50%;
+        top: 50%;
+        transform: translate(-50%, -50%);
+
+        font-size: 30px;
+        font-family: Arial, Helvetica, sans-serif;
```
anstsot: better to add font styles to `html` selector

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+
+    &:hover {
+      cursor: pointer;
```
anstsot: this style should be added to `.calendar__day` without `hover` state

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+
+    transition-duration: 0.5s;
+
+    @for $box from 1 through $box-col {
+      &:nth-child(#{$box})::before {
+        content: "#{$box}";
+        position: absolute;
+        left: 50%;
+        top: 50%;
+        transform: translate(-50%, -50%);
```
anstsot: why do you need absolute positioning? maybe you can add needed padding?

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.
And please fix the bug when calendar has wider gaps in case of 28 days 
image

src/styles/blocks/calendar.scss
MariiaMonakhova: Thank you for your feedback

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  gap: 1px;
+
+  padding: 10px;
+  margin: 0 auto;
+  padding-top: 50px;
+
+  $days: 0 "mon", 1 "tue", 2 "wed", 3 "thu", 4 "fri", 5 "sat", 6 "sun";
```
tetiana-ivanchenko: move it to the variables file

tetiana-ivanchenko (approved):

PR #2277, opened 2023-06-01
Add Calendar Solution

NataliGru:
- [DEMO LINK](https://NataliGru.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -17,20 +17,20 @@
   @for $length from 28 through 31 {
     &--month-length--#{$length} {
       .calendar__day:not(:nth-child(-n + #{$length})) {
-        visibility: hidden;
+        display: none;
       }
     }
   }
 
   &__day {
-    position: relative;
     box-sizing: border-box;
     display: flex;
     justify-content: center;
     align-items: center;
+
     width: $calendar-cell-side;
     height: $calendar-cell-side;
-    
+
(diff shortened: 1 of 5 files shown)


tetiana-ivanchenko (changes requested):
provide the demo link, also check all points from the checklist (7)

src/styles/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    &--start-day--#{$day} {
+      .calendar__day:first-child {
+        margin-left: firstDayShift($num, $shift);
+      }
+    }
+  }
+
+  @for $length from 28 through 31 {
+    &--month-length--#{$length} {
+      .calendar__day:not(:nth-child(-n + #{$length})) {
```
juljaVysotska: Too complicated selector. You said `for all elements before #{$length} don not do display: none`, when you can said `for all elements after(!!!) #{$length} do display: none`. Think how we can make selector easier.

src/styles/functions.scss
```diff
@@ -0,0 +1,7 @@
+@function getWidth($side, $gap) {
```
juljaVysotska: Goood🔥🔥🔥

juljaVysotska (changes requested):
Good job! Please, check and fix issues.

Best regards🌸

tetiana-ivanchenko (approved):

PR #2286, opened 2023-06-02
add task solution

ArtemAndrushchenko:
[DEMO LINK](https://ArtemAndrushchenko.github.io/layout_calendar/)


tetiana-ivanchenko (changes requested):
your solution doesn't work
image

src/index.html
juljaVysotska, line 1: You should display a calendar in the middle of the screen. It's task.

src/styles/main.scss
```diff
@@ -1,3 +1,77 @@
+$week: (
```
juljaVysotska: A good idea will be to store variables in a separate file.

src/styles/calendar.scss
juljaVysotska, line 1: This file is not linked into HTML. Do you really need this file?

src/styles/main.scss
```diff
@@ -1,3 +1,77 @@
...
+);
+
+$day-width: 100px;
+$day-height: 100px;
+$gap: 1px;
+$padding: 10px;
+$background-color-day: #eee;
+$background-color-day-active:pink;
+
 body {
```
juljaVysotska: You can move reset properties into a separate file.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them.

Best regards 🌸

tetiana-ivanchenko (changes requested):
image

please, before re-requesting review make sure that your solution works as expected (modificator of week day doesn't work)

tetiana-ivanchenko (approved):

PR #2288, opened 2023-06-02
task Calendar solution

outofs:
https://outofs.github.io/layout_calendar/
https://outofs.github.io/layout_calendar/report/html_report/


tetiana-ivanchenko (approved):
GJ

PR #2289, opened 2023-06-02
Solution added

kyrylomanko:
- [DEMO LINK](https://kyrylomanko.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar calendar--month-length-28 calendar--start-day-wed">
+    <div class="calendar calendar--month-length-29 calendar--start-day-mon">
       <div class="calendar__cell"></div>
       <div class="calendar__cell"></div>
       <div class="calendar__cell"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--month-length-28 calendar--start-day-wed">
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+      <div class="calendar__cell"></div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,2 @@
+$block-size: 100px;
```
Anastasiia-Svintsova: Consider renaming the variable '$block-size' to something more descriptive, like '$calendar-cell-size', to improve code readability.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  position: relative;
+  max-width: 708px;
```
Anastasiia-Svintsova: The max-width should be calculated based on the block size, gap, and paddings to make it more flexible.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  margin: 40px auto;
+  padding: 10px;
+
+  &__cell {
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
+    text-align: center;
+
+    display: block;
```
Anastasiia-Svintsova: `div` has a `display: block;` by default, so remove it

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--month-length-28 calendar--start-day-wed">
```
Anastasiia-Svintsova: - Modifiers are not working. Fix it, please 
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  position: relative;
+  max-width: 708px;
+  margin: 40px auto;
```
Anastasiia-Svintsova: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  position: relative;
```
Anastasiia-Svintsova: Are you sure you need this position property? It seems redundant

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__cell {
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
+    text-align: center;
+
+    display: block;
+    width: $block-size;
+    height: $block-size;
+    line-height: $block-size;
```
Anastasiia-Svintsova: Using line-height for alignment can lead to issues and is generally not considered the best practice.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+    background-color: #eee;
+    border: 1px black solid;
+
+    transition-duration: 0.5s;
```
Anastasiia-Svintsova, line 25: Specify a transition property as well

Anastasiia-Svintsova (changes requested):
Good start! Pay attention to my comments, please

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+    $calendar-column-gap,
+    $calendar-padding
+  );
+  padding: $calendar-padding;
+
+  &__cell {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    font-family: Arial, Helvetica, sans-serif;
```
Anastasiia-Svintsova, line 16: It's better to move font-family to a body tag

src/styles/main.scss
```diff
@@ -1,3 +1,16 @@
...
+@import "./blocks/calendar.scss";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
Anastasiia-Svintsova, line 13: It's not the best idea to add a display: flex to the body tag. Would be better to use class name here

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,4 @@
+@function width($calendar-cell-size, $calendar-column-gap, $calendar-padding) {
```
Anastasiia-Svintsova, line 1: Great idea to use a function here

Anastasiia-Svintsova (approved):
Good job

PR #2311, opened 2023-06-05
Develop

mileantkostya:
https://mileantkostya.github.io/layout_calendar/

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,64 @@
+.calendar {
+  width: 7.1 * $box-width;
+  height: 5.1 * $box-height;
+
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+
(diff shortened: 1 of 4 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+        content: "#{$box}";
+        position: absolute;
+        left: 50%;
+        top: 50%;
+        transform: translate(-50%, -50%);
+
+        font-size: 30px;
+        font-family: Arial, Helvetica, sans-serif;
+      }
+    }
```
tetiana-ivanchenko: checklist:
Don't use SASS loops for styles that stay the same for all elements of the group, e.g. display or position.

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
image

PR #2332, opened 2023-06-09
add calendar

daria-sazon:
- [DEMO LINK](https://daria-sazon.github.io/layout_calendar/)


etojeDenys (changes requested):
i should have the opportunity to change the number of cells just by changing the modifier
image

etojeDenys (changes requested):
still not fixed
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  gap: $gap-size;
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
```
OleziO, line 11: Move to the * selector and remove it from all other places.

OleziO (changes requested):
image
This number should set days count

daria-sazon:
@etojeDenys @OleziO I don't understand why but it works for me 
image

demetrxx (approved):
Well done 👍 Go ahead!

PR #2334, opened 2023-06-11
add task solution

yslyshchenko:
[DEMO LINK](https://yslyshchenko.github.io/layout_calendar/)
Test results weren't opened in a browser, so there is no link.

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -15,13 +15,13 @@
   padding: 10px;
 
   @each $day, $index in $day-of-the-week {
-    &--start-day--#{$day}:first-child {
+    &--start-day--#{$day} :first-child {
       margin-left: ($cell-size + $cell-indent) * $index;
     }
   }
 
   @for $last-day from 28 through 31 {
-    &--month-length--#{$last-day}:nth-child(n + #{$last-day}) {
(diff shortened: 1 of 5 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
i set tue and 28 modificators, but it doesn't work 
image

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,10 @@
+$cell-size: 100px;
```
volodymyr-soltys97, line 1: Consider using a more descriptive variable name instead of $cell-size, such as $calendar-cell-size.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,10 @@
...
+$cell-indent: 1px;
```
volodymyr-soltys97, line 2: Consider renaming $cell-indent to $calendar-cell-gap to better represent its purpose in the context of the task.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+.container {
+  display: flex;
```
volodymyr-soltys97, line 2: Change .container class to a BEM-styled name, like .calendar-container.

volodymyr-soltys97 (approved):
Good job 👍

PR #2348, opened 2023-06-14
add task solution

Phoenix0980:


Phoenix0980:
https://Phoenix0980.github.io/layout_calendar/

nazarmatsevych (changes requested):
Good start, keep going

volodymyr-soltys97 (changes requested):
The upper block is not located in line with the others
Also you need to add modifier for month length
image

OleziO (changes requested):
Move the calendar to the center of the screen to avoid this bug
image

image
Amount of days must depend of the class number

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+  $days: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
+
+  @each $day, $value in $days {
+    &--start-day-#{$day} &__day:first-child {
+      margin-left: ($cell-size + $gap) * $value;
+    }
+  }
+
+  @for $lenght from 28 through 31 {
+    .calendar—month-length-#{$lenght} &__day:nth-child(n + #{$lenght + 1}) {
```
anastasiiavorobiova: You should hide the cells after the specified length, not modify their content. Use 'display: none;' instead of setting the content property.

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
   margin: 0;
 }
+
+* {
+  box-sizing: border-box;
+}
+
+.calendar {
+  gap: $gap;
+  width: 750px;
```
anastasiiavorobiova: The width of the calendar should be calculated based on the cell size, gap, and padding. Use the formula 'width: ($cell-size * 7) + ($gap * 6) + 10px;'

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+
+  $days: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
+
+  @each $day, $value in $days {
+    &--start-day-#{$day} &__day:first-child {
+      margin-left: ($cell-size + $gap) * $value;
+    }
+  }
+
+  @for $lenght from 28 through 31 {
```
anastasiiavorobiova: There is a typo in the variable name, change '$lenght' to '$length'.

anastasiiavorobiova (changes requested):
Great work! Let's make it perfect! Days quantity modifier doesn't work well

src/styles/main.scss
```diff
@@ -1,3 +1,65 @@
...
+    width: 100px;
+    height: 100px;
+    border: 1px solid black;
+  }
+
+  &__day:hover {
+    background-color: pink;
+    cursor: pointer;
+    transform: translateY(-20px);
+    transition-duration: 500ms;
```
volodymyr-soltys97: You can rewrite it like this
```suggestion
  &__day {
    background-color: #eee;
    width: 100px;
    height: 100px;
    border: 1px solid black;
    transition-duration: 500ms;
    
    &:hover {
      background-color: pink;
      cursor: pointer;
      transform: translateY(-20px);
      transition-duration: 500ms; 
    }
  }
```

volodymyr-soltys97 (changes requested):
The number of days does not work in the modifier
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,35 @@
...
+    width: 100px;
+    height: 100px;
+    border: 1px solid black;
+    transition-duration: 500ms;
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+      transition-duration: 500ms;
```
TarasHoliuk, line 20: redundant since you set it for `&__day` (should be only there)
Also, it's a good practice to set proper `transition-property`

Just FYI

TarasHoliuk (approved):
GJ! Approved :heavy_check_mark: 
Added 1 comment FYI - no need to re-request review

PR #2361, opened 2023-06-21
calendar

Stako21:


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,16 @@
+* {
+  display: flex;
```
nazarmatsevych, line 2: Use * selector only for box-sizing rule

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+}
+
+$startDays: (
+  "mon": 0,
+  "tue": 101px,
+  "wed": 202px,
+  "thu": 303px,
+  "fri": 404px,
+  "sat": 505px,
+  "sun": 606px,
```
nazarmatsevych, line 43: It's better to calculate it inside loop

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
+.calendar {
+  width: 726px;
```
nazarmatsevych, line 2: it's better to calculate this width using sass function

nazarmatsevych (approved):
LGTM

PR #2404, opened 2023-07-06
Add task solution

klaudia-klis:
- [DEMO LINK](https://klaudia-klis.github.io/layout_calendar/)


src/index.html
```diff
@@ -7,6 +7,38 @@
...
+      <div class="calendar__day calendar__day--22"></div>
+      <div class="calendar__day calendar__day--23"></div>
+      <div class="calendar__day calendar__day--24"></div>
+      <div class="calendar__day calendar__day--25"></div>
+      <div class="calendar__day calendar__day--26"></div>
+      <div class="calendar__day calendar__day--27"></div>
+      <div class="calendar__day calendar__day--28"></div>
+      <div class="calendar__day calendar__day--29"></div>
+      <div class="calendar__day calendar__day--30"></div>
+      <div class="calendar__day calendar__day--31"></div>
```
mvjl000: Those divs need only one class - in this case calendar__day. Get rid of the other one, you still will be able to solve it that way ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  padding: 10px;
+  gap: $daysGap;
+  width: calc(7*100px + 7*1px);
```
mvjl000: To calculate width, use your variables

also, you need one adjustment - 7 day cells + 6 gaps

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &__day {
+    box-sizing: border-box;
+    background-color: #eee;
+    width: $dimensions;
+    height: $dimensions;
+    text-align: center;
+    line-height: $dimensions;
+    border: 1px solid black;
+    transition: transform 0.5s;
```
mvjl000: List all properties that you apply transition to. It will be easier to understand which properties are affected by it and also there will be no case in which something that should not have a transition has one

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+
+    &:hover {
+      cursor: pointer;
```
mvjl000: I would move it up so it's directly inside of __day

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+
+  @for $day from 1 through 31 {
+    &__day--#{$day}::before {
+      content: '#{$day}';
+      font-size: 30px;
+      font-family: Arial, Helvetica, sans-serif;
```
mvjl000: Inside of loops, set only properties that are changing. In this case content is fine but font styles should be inside of __day

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  $days: (
+    "mon": 0,
+    "tue": calc(1*#{$dimensions} + 1*#{$daysGap}),
+    "wed": calc(2*#{$dimensions} + 2*#{$daysGap}),
+    "thu": calc(3*#{$dimensions} + 3*#{$daysGap}),
+    "fri": calc(4*#{$dimensions} + 4*#{$daysGap}),
+    "sat": calc(5*#{$dimensions} + 5*#{$daysGap}),
+    "sun": calc(6*#{$dimensions} + 6*#{$daysGap})
```
mvjl000: Mayber instead of calculating the margin for each day here, simply put the index of each day and calculate it inside @each with help of other variables.

Btw, this could be inside of variables file

mvjl000 (changes requested):

choeqq (approved):
All good! Really nice 🚀

PR #2408, opened 2023-07-07
task solution

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,47 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  gap: $calendar-gap;
+  align-items: center;
+  padding: $calendar-padding;
+  margin:0 auto;
+  padding-top: 100px;
+
+  &__field {
+    @include field($size);
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY($field-translate-Y);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+      }
+    }
+  }
+
+  $start-day:
+    'mon' 0,
+    'tue' 1,
+    'wed' 2,
+    'thu' 3,
+    'fri' 4,
+    'sat' 5,
+    'sun' 6;
+
+  @each $day,$i in $start-day {
+    &__start-day--#{$day} > :first-child {
+      margin-left: $size * $i + $calendar-gap * $i;
+    }
+  }
+
+  @for $i from 28 to 31 {
+    &__month-length--#{$i} > :nth-child(n + #{$i + 1}) {
+      display: none;
+    }
+  }
+}
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -3,16 +3,20 @@
   flex-wrap: wrap;
   width: $calendar-width;
   gap: $calendar-gap;
-  align-items: center;
-  padding: $calendar-padding;
-  margin:0 auto;
-  padding-top: 100px;
 
   &__field {
     @include field($size);
+    box-sizing: border-box;
+    border: $field-border solid black;
+    background-color: #eee;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    transition-duration: 0.5s;
+    transition-property: transform;
+    cursor: pointer;
 
     &:hover {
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,7 @@
+$size: 100px;
+$calendar-gap: 1px;
+$calendar-width: ($size + $calendar-gap) * 7;
```
mvjl000: Almost ;)
7 columns equals 6 gaps. Please correct this.

7 fiels + 6 gaps

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,7 @@
...
+$calendar-padding: 10px;
+$font-size: 30px;
+$field-border: 1px;
+$field-translate-Y: -20px;
```
mvjl000, line 7: There is no need to create variables for values that you use only once

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,11 @@
...
+  width: $size;
+  height: $size;
+  box-sizing: border-box;
+  border: $field-border solid black;
+  background-color: #eee;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition-duration: 0.5s;
+}
```
mvjl000: This mixin is unnecessary here. All of these properties should be inside of __field in a regular way.

If you really want to use a mixin here, I think this would make more sense:
```suggestion
@mixin dimensions($size) {
  width: $size;
  height: $size;
}
```

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,11 @@
+@mixin field($size) {
+  width: $size;
+  height: $size;
+  box-sizing: border-box;
+  border: $field-border solid black;
+  background-color: #eee;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition-duration: 0.5s;
```
mvjl000: - Use transition property here
- List all properties that you apply transition to. It will be easier to understand which properties are affected by it and also there will be no case in which something that should not have a transition has one

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  $start-day:
+    'mon' 0,
+    'tue' 1,
+    'wed' 2,
+    'thu' 3,
+    'fri' 4,
+    'sat' 5,
+    'sun' 6;
```
mvjl000: This could go to variables file

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  align-items: center;
+  padding: $calendar-padding;
+  margin:0 auto;
+  padding-top: 100px;
+
+  &__field {
+    @include field($size);
+
+    &:hover {
+      cursor: pointer;
```
mvjl000: I would prefer it to be directly inside of __field, not :hover

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  gap: $calendar-gap;
+  align-items: center;
+  padding: $calendar-padding;
+  margin:0 auto;
+  padding-top: 100px;
```
mvjl000: The calendar should be centered both horizontally and vertically. Get rid of all margins and paddings here and simply center it by adding flex to body (don't forget about justify, align and height)

mvjl000 (changes requested):

src/styles/utils/_variables.scss
Sancz0pansa: Understood

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+  &__field {
+    @include field($size);
+    box-sizing: border-box;
+    border: $field-border solid black;
+    background-color: #eee;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    transition-duration: 0.5s;
+    transition-property: transform;
```
choeqq, line 16: you can also use transition - it is property that combines all possible `transition-something` properties 😉 

[Transition](https://www.w3schools.com/css/css3_transitions.asp)

choeqq (approved):
All good, please take a look at my comment and you're good to go! 🚀

PR #2409, opened 2023-07-07
Develop

Meg-Sowka:
- [DEMO LINK](https://Meg-Sowka.github.io/layout_calendar/)
--- no tests for this task ---

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -3,18 +3,6 @@
   flex-wrap: wrap;
   gap: $calendarGap;
   max-width: $calendarSize;
-  padding: $calendarPadding;
-  margin-left: auto;
-  margin-right: auto;
-
-  $days:
-    "mon" 1,
-    "tue" 2,
-    "wed" 3,
-    "thu" 4,
-    "fri" 5,
-    "sat" 6,
-    "sun" 7;
 
   @each $name, $dayNr in $days {
     &--start-day-#{$name} &__day:first-child {
@@ -31,23 +19,16 @@
   &__day {
     width: $daySize;
     height: $daySize;
+    display: flex;
+    align-items: center;
+    justify-content: center;
     border: 1px solid black;
     background-color: $calendarBg;
-    position: relative;
-    transition: all 0.5s ease-in;
+    transition: background-color 0.5s ease-in, transform 0.5s ease-in;
 
     &::before {
-      content: " ";
       font-family: Arial, Helvetica, sans-serif;
       font-size: 30px;
-      display: flex;
-      position: absolute;
-      left: 0;
-      top: 0;
-      width: 100%;
-      height: 100%;
-      align-items: center;
-      justify-content: center;
     }
 
     @for $i from 1 through 31 {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,65 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendarGap;
+  max-width: $calendarSize;
+  padding: $calendarPadding;
+  margin-left: auto;
+  margin-right: auto;
```
mvjl000: The calednar should be centered both horizontally and vertically. You could get rid of these margins nad paddings and simply add flex to body (remember about justify/align and height) ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,65 @@
...
+  &__day {
+    width: $daySize;
+    height: $daySize;
+    border: 1px solid black;
+    background-color: $calendarBg;
+    position: relative;
+    transition: all 0.5s ease-in;
+
+    &::before {
+      content: " ";
```
mvjl000: Since you are setting the content in the @for loop, you don't need it here

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,65 @@
...
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      display: flex;
+      position: absolute;
+      left: 0;
+      top: 0;
+      width: 100%;
+      height: 100%;
```
mvjl000: Instead of positioning it absolutely, try to center it with flex added to __day.
Your solution works but it can be simpler ;)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$daySize: 100px;
+$calendarGap: 1px;
+$calendarPadding: 10px;
+$calendarSize: ($daySize * 7) +  ($calendarGap * 6) + ($calendarPadding * 2);
```
mvjl000: Good calculations but you don't need that padding in your calendar.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  &__day {
+    width: $daySize;
+    height: $daySize;
+    border: 1px solid black;
+    background-color: $calendarBg;
+    position: relative;
+    transition: all 0.5s ease-in;
```
mvjl000: List all properties that you apply transition to. It will be easier to understand which properties are affected by it and also there will be no case in which something that should not have a transition has one

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,65 @@
...
+  margin-right: auto;
+
+  $days:
+    "mon" 1,
+    "tue" 2,
+    "wed" 3,
+    "thu" 4,
+    "fri" 5,
+    "sat" 6,
+    "sun" 7;
```
mvjl000: I think this can go to variables file

mvjl000 (changes requested):
Some minor changes needed but you nailed all the hard parts - keep it up!

asakevych (approved):
Great job!! 🚀 🔥 🔥

PR #2410, opened 2023-07-07
add task solution

ogospadzipalina:
[DEMO LINK](https://ogospadzipalina.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
+@import "../utils/variables";
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: 706px;
```
mvjl000: This 706 is like a magic number, it is not clear why is it set to 706px and if you change for instance dimensions of calendar-day things will break up.

src/index.html
```diff
@@ -7,6 +7,40 @@
...
+        <div class="calendar__day-box calendar__day-box--22"></div>
+        <div class="calendar__day-box calendar__day-box--23"></div>
+        <div class="calendar__day-box calendar__day-box--24"></div>
+        <div class="calendar__day-box calendar__day-box--25"></div>
+        <div class="calendar__day-box calendar__day-box--26"></div>
+        <div class="calendar__day-box calendar__day-box--27"></div>
+        <div class="calendar__day-box calendar__day-box--28"></div>
+        <div class="calendar__day-box calendar__day-box--29"></div>
+        <div class="calendar__day-box calendar__day-box--30"></div>
+        <div class="calendar__day-box calendar__day-box--31"></div>
```
mvjl000: Those divs should have only one class name - calendar__day-box, get rid of the other one. You can solve it without it

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+  padding: 10px;
+  gap: $gap-size;
+  margin: 0 auto;
```
mvjl000: since you center it properly with the help of body, you don't need margin and padding

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$box-size: 100px;
+$gap-size: 1px;
+$calendar-size: $box-size * 7 + $gap-size * 6;
```
mvjl000, line 3: Make use of this variable in calendar width

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+    box-sizing: border-box;
+    text-align: center;
+    line-height: $box-size;
+    transition: transform 0.5s;
+
+    @for $day from 1 through 31 {
+      &--#{$day}::before {
+        content: "#{$day}";
+        font-size: 30px;
+        font-family: Arial, Helvetica, sans-serif;
```
mvjl000: Inside of loop, put only the properties that change (content). Font styles should be up in calendar box

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
```
mvjl000: Move the `cursor: pointer;` up to calendar box

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+
+  .calendar__day-box {
+    background-color: #eee;
+    width: $box-size;
+    height: $box-size;
+    border: 1px solid black;
+    box-sizing: border-box;
+    text-align: center;
+    line-height: $box-size;
+    transition: transform 0.5s;
```
mvjl000: You should also list background color here because currently it's changing to pink immediately

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  $days:
+    "mon" 0,
+    "tue" 1,
+    "wed" 2,
+    "thu" 3,
+    "fri" 4,
+    "sat" 5,
+    "sun" 6;
```
mvjl000: This could be in variables file

mvjl000 (changes requested):

src/styles/blocks/calendar.scss
ogospadzipalina: i even had it in variables but never used it 🤦

src/styles/blocks/calendar.scss
ogospadzipalina: Thank you for your comments! all done

asakevych (approved):
Great job! 🚀 🔥 🔥

PR #2412, opened 2023-07-07
Calendar - develop branch

marekder:
- [DEMO LINK](https://marekder.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  &__day {
+    width: $day-size;
+    height: $day-size;
+    border: 1px solid black;
+    background-color: #eee;
+    position: relative;
+
+    &::before {
+      content: counter(day);
```
mvjl000: According to the checklist, you should set the value of before with help of [@for](https://sass-lang.com/documentation/at-rules/control/for/).

Try to implement it that way and get rid of all CSS counters.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+.calendar {
```
mvjl000: Calendar should be in the middle of the screen. Try to center it ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+    background-color: #eee;
+    position: relative;
+
+    &::before {
+      content: counter(day);
+      counter-increment: day;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
```
mvjl000: You could simplify it. Instead of positioning it absolute just add display flex to `&__day`.
And don't forget there about `justify-content` and `align-items` ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+      font-family: Arial, sans-serif;
+      font-size: $day-font-size;
+    }
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+      transition: all 0.5s;
```
mvjl000: I would place `cursor: pointer;` inside of `&__day`.
Same thing with transition, it should be there because right now, the animation is broken. Move it up and you will se nice and smooth animation.

So, `&:hover` should look like this:
```suggestion
      background-color: pink;
      transform: translateY(-20px);
```

mvjl000: Btw avoid setting transition to `all`, list all styles that you apply transition to. It's more readable and you know exactly what is affected by it.

mvjl000 (changes requested):

asakevych (approved):
Great job!! 🚀 👏 👏

PR #2414, opened 2023-07-07
add task solution

HaniaNassalska:
https://hanianassalska.github.io/layout_calendar/

### DIFF
--- a/src/styles/blocks/caledar.scss
+++ b/src/styles/blocks/caledar.scss
@@ -31,7 +31,7 @@
     background-color: $grey;
     font-family: Arial, sans-serif;
     font-size: 30px;
-    transition-duration: 0.5s;
+    transition: background-color 0.5s, transform 0.5s;
 
     &:hover {
       background-color: pink;
--- a/src/styles/blocks/caledar.scss
+++ b/src/styles/blocks/caledar.scss
@@ -5,17 +5,8 @@
   gap: $calendar-gap;
   padding: $calendar-padding;
 
-  $days: (
-    0 "mon",
-    1 "tue",
-    2 "wed",
-    3 "thu",
-    4 "fri",
-    5 "sat",
-    6 "sun"
-  );
-
-  @each $number, $day in $days {
+  @each $number,
+  $day in $days {
     &--start-day-#{$day} :first-child {
       margin-left: $number * $calendar-day-space;
     }
@@ -41,19 +32,17 @@
     transition: 0.5s;
     font-family: Arial, sans-serif;
     font-size: 30px;
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+    }
   }
 
   @for $index from 1 through 31 {
     &__day:nth-child(#{$index})::before {
       content: "#{$index}";
-    }
-  }
-
-  @for $index from 1 through 31 {
-    &__day:nth-child(#{$index}):hover {
-      background-color: pink;
-      cursor: pointer;
-      transform: translateY(-20px);
     }
   }
 }
--- a/src/styles/blocks/caledar.scss
+++ b/src/styles/blocks/caledar.scss
@@ -1,0 +1,69 @@
+$grey: #eee;
+$squere-side: 100px;
+$calendar-gap: 1px;
+$calendar-day-border: 1px;
+$calendar-day-space: $squere-side + $calendar-gap;
+$calendar-padding: 10px;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,69 @@
+$grey: #eee;
+$squere-side: 100px;
+$calendar-gap: 1px;
+$calendar-day-border: 1px;
+$calendar-day-space: $squere-side + $calendar-gap;
+$calendar-padding: 10px;
+$calendar-widht: ($squere-side + $calendar-gap) * 7;
```
mvjl000: Please move variables to a separate file

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.calendar {
+  width: $calendar-widht;
+  margin: 0 auto;
```
mvjl000: The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,69 @@
...
+    justify-content: center;
+    align-items: center;
+    transition: 0.5s;
+  }
+
+  @for $index from 1 through 31 {
+    &__day:nth-child(#{$index})::before {
+      content: "#{$index}";
+      font-family: Arial, sans-serif;
+      font-size: 30px;
```
mvjl000: This should be inside of day. Inside of loops, put only the properties that are relying on the loop - in this case content.

mvjl000 (changes requested):
Some minor changes needed but besides of that it looks good.
Keep it up!

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  $days: (
+    0 "mon",
+    1 "tue",
+    2 "wed",
+    3 "thu",
+    4 "fri",
+    5 "sat",
+    6 "sun"
+  );
```
choeqq: that also belongs to variables file

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  @for $index from 1 through 31 {
+    &__day:nth-child(#{$index}):hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+    }
+  }
```
choeqq: those styles aren't really changing - please apply them to `__day`

choeqq (changes requested):
Almost there, little changes needed! 🚀

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,48 @@
...
+  &__day {
+    box-sizing: border-box;
+    background-color: $grey;
+    width: $squere-side;
+    height: $squere-side;
+    border: $calendar-day-border solid black;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: 0.5s;
```
mvjl000: [Check list](https://github.com/mate-academy/layout_calendar/blob/master/checklist.md) point 8th: Make sure to list all styles that you apply transition to (don't use `all` either)

mvjl000 (changes requested):
One more small change - checklist requirements, sorry that I didn't notice it earlier.
And one more thing from the checklist (12), I would like you to [reorder your properties ](https://github.com/necolas/idiomatic-css#declaration-order)so it's cleaner ;)

src/styles/blocks/caledar.scss
```diff
@@ -0,0 +1,48 @@
...
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    width: $squere-side;
+    height: $squere-side;
+    border: $calendar-day-border solid black;
+    background-color: $grey;
+    font-family: Arial, sans-serif;
+    font-size: 30px;
+    transition-duration: 0.5s;
```
asakevych: something is wrong with transition here: 
1. you just set transition-duration, but without any property
2. you can use transition like here: https://www.w3schools.com/css/css3_transitions.asp
```suggestion
    transition: background-color 0.5s, transform 0.5s;
```
or 
```suggestion
    transition-property:  background-color, transform;
    transition-duration: 0.5s;
```

asakevych (changes requested):
Nearly perfect!! just one small change left 😉

choeqq (approved):
Looks good! Approved ✅

PR #2415, opened 2023-07-07
add task solution-no1

cardinal1991:
- [DEMO LINK](https://cardinal1991.github.io/layout_calendar/)
- [TEST REPORT LINK](https://cardinal1991.github.io/layout_calendar/report/html_report/)/empty


src/styles/main.scss
```diff
@@ -1,3 +1,9 @@
+@import "./blocks/root.scss";
+@import "./blocks/calendar.scss";
+
 body {
   margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
```
mvjl000: The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body.
Your approach is good but the one thing that you are missing is height set to full screen (100vh)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: 1px;
+  margin: auto;
+  width: ($cellsize * 7) + 6px;
```
mvjl000: I appreciate that you are calculating the width basing on cellsize but maybe there is a way to make it more understandable / readable.
I know that 6px refers to gaps but make it more obvious, using your gap variable

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+
+.calendar {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: 1px;
```
mvjl000: gap variable should be used here

src/styles/blocks/calendar.scss
mvjl000, line 1: Important thing for your loops:
Inside of loops put only the styles that are changing. Everything else that stays the same should be outside the loop

src/styles/blocks/root.scss
```diff
@@ -0,0 +1,12 @@
+$cellsize: 100px;
+$cellgap : 1px;
+
+$days: (6: "-mon", 5: "-tue", 4: "-wed", 3: "-thu",
+2: "-fri", 1: "-sat", 0: "-sun");
```
mvjl000: Why is it backwords?
Go for mon: 0, tue: 1 etc... it will be way cleaner

mvjl000 (changes requested):
You need to make some changes. Especially to first item's margin logic because it's not working correctly

src/styles/blocks/root.scss
cardinal1991: Its small margin-left counting machine. In one of the modificator, theoreticly we should do margin left for the first child, that depends on which day we have. So we can do it like with a pairs map. The number is depending on how many days we should "remove" from the first line, depending on which day(monday) we are starting month. So if monday is 1st day we should remowe 6 another days cell from the first line, and according to the questions we are setting it by margin left.

src/styles/blocks/calendar.scss
cardinal1991: that was way to solve problem with
 41:7  error  Tag can't have more than one element  [calendar__cell]     (one-element)
 41:7  error  Tag can't have more than one element  [calendar__cell-31]  (one-element)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+}
+
+@for $i from 1 through 31 {
+  .calendar__cell-#{$i}::before {
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
+    content: "#{$i}" ;
+    padding: auto;
```
choeqq: only `content` property is changing here - please move unchanged properieties to cell

choeqq (changes requested):
1 little change and you're good to go! 🚀

choeqq (approved):
All good, nicely done 🚀

PR #2420, opened 2023-07-08
solution

Vira-V:
- [DEMO LINK](https://Vira-V.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -51,7 +51,7 @@ $cellSize: 100px;
   @each $day, $index in $monthDays {
     &--start-day-#{$day} &__day {
       &:first-child {
-        margin-left: ($cellSize + $gap * 2.7) * $index;
+        margin-left: ($cellSize + $gap * 2.75) * $index;
       }
     }
   }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,28 +1,31 @@
-$padding: 10px;
-$gap: 1px;
-$columns: 7;
-$cellSize: 100px;
+@import "./utils/variables";
+
+body {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
 
 .calendar {
   display: flex;
   flex-wrap: wrap;
-  justify-content: flex-start;
   padding: $padding;
-  max-width: ($cellSize + $gap) * ($columns + 1);
-  margin: 0 auto;
+  width: $cellSize * 7 + 100 + ($gap * 6);
   gap: $gap;
 
   &__day {
+    display: flex;
+    align-items: center;
+    justify-content: center;
     width: $cellSize;
     height: $cellSize;
-    background-color: #eee;
+    background-color: $cell-background-color;
     border: 1px solid black;
-    position: relative;
-    transition: all ease-in-out 0.3s;
+    transition: ease-in-out 0.3s;
 
     &:hover {
       cursor: pointer;
-      background-color: pink;
+      background-color: $hover-background-color;
       transform: translateY(-20px);
     }
 
@@ -30,23 +33,14 @@ $cellSize: 100px;
       content: "";
       font-family: Arial, Helvetica, sans-serif;
       font-size: 30px;
-      position: absolute;
-      top: 50%;
-      left: 50%;
-      transform: translate(-50%, -50%);
-      text-align: center;
+    }
+
+    @for $index from 1 through 31 {
+      &:nth-child(#{$index})::before {
+        content: "#{$index}";
+      }
     }
   }
-
-  $monthDays: (
-    "sun": 0,
-    "mon": 1,
-    "tue": 2,
-    "wed": 3,
-    "thu": 4,
-    "fri": 5,
-    "sat": 6,
-  );
 
   @each $day, $index in $monthDays {
     &--start-day-#{$day} &__day {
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
-body {
-  margin: 0;
+$padding: 10px;
+$gap: 1px;
+$columns: 7;
+$cellSize: 100px;
```
mvjl000: Move all variables to separate file

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  padding: $padding;
+  max-width: ($cellSize + $gap) * ($columns + 1);
```
mvjl000: You need to adjust (and simplify) your calculations:

- change this property to width
- you need 7 cellSize and 6 gaps

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+  margin: 0 auto;
+  gap: $gap;
+
+  &__day {
+    width: $cellSize;
+    height: $cellSize;
+    background-color: #eee;
+    border: 1px solid black;
+    position: relative;
+    transition: all ease-in-out 0.3s;
```
mvjl000: List all properties that you apply transition to. It will be easier to understand which properties are affected by it and also there will be no case in which something that should not have a transition has one

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+
+  $monthDays: (
+    "sun": 0,
+    "mon": 1,
+    "tue": 2,
+    "wed": 3,
+    "thu": 4,
+    "fri": 5,
+    "sat": 6,
+  );
```
mvjl000: Let's move it to variables file

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+
+  @each $day, $index in $monthDays {
+    &--start-day-#{$day} &__day {
+      &:first-child {
+        margin-left: ($cellSize + $gap * 2.75) * $index;
+      }
+    }
+  }
+
+  $monthLength: 28, 29, 30, 31;
```
mvjl000: You musn't use variable for this. And I suppose that's the reason your solution breaks after changing modificator to different value than 28,29,30,31. When changed to for example `abc` all of your calendar text disappears. Try to fix it.

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+    &--month-length-31 .calendar__day:nth-child(#{$i})::before,
+    &--month-length-30
+    .calendar__day:nth-child(-n + 30):nth-child(#{$i})::before,
+    &--month-length-29
+    .calendar__day:nth-child(-n + 29):nth-child(#{$i})::before,
+    &--month-length-28
+    .calendar__day:nth-child(-n + 28):nth-child(#{$i})::before {
+      content: "#{$i}";
+    }
+  }
```
mvjl000: You can make it way simpler. As a hint I can tell you that you should do this loop inside of `__day` and you need only one selector inside of it

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+$columns: 7;
+$cellSize: 100px;
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  padding: $padding;
+  max-width: ($cellSize + $gap) * ($columns + 1);
+  margin: 0 auto;
```
mvjl000: The calendar should be centered both horizontally and vertically. Simply add flex to body (no need for margin here)

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+
+    &::before {
+      content: "";
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      text-align: center;
```
mvjl000: Instead of positioning it absolutely do a less tricky approach - center it with flex in __day

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+
+  $monthLength: 28, 29, 30, 31;
+
+  @each $length in $monthLength {
+    &--month-length-#{$length} &__day {
+      &:nth-child(n + #{$length + 1}) {
+        display: none;
+      }
+    }
+  }
```
mvjl000: Try to do it with @for - you should care about range 28-31

mvjl000 (changes requested):

Vira-V:
changes uploaded

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  padding: $padding;
+  width: $cellSize * 7 + 100 + ($gap * 6);
```
asakevych: you shouldn't add 100px here, just (`$cellSize * 7) + ($gap * 6);`

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+      &:nth-child(#{$index})::before {
+        content: "#{$index}";
+      }
+    }
+  }
+
+  @each $day, $index in $monthDays {
+    &--start-day-#{$day} &__day {
+      &:first-child {
+        margin-left: ($cellSize + $gap * 2.75) * $index;
```
asakevych: it's also not the best solution- why do you need to multiply by 2.75?
it seems it should be `($cellSize + $gap)  *  $index `

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
+@import "./utils/variables";
+
 body {
-  margin: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
```
asakevych: add `box-sizing: border-box` for all elements, using `*` selector

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+.calendar {
```
asakevych: every block should be in seperate file, create blocks/calendar.scss file with those styles

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+
+  &__day {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: $cellSize;
+    height: $cellSize;
+    background-color: $cell-background-color;
+    border: 1px solid black;
+    transition: ease-in-out 0.3s;
```
asakevych: list all properties like is in example here: https://www.w3schools.com/css/css3_transitions.asp
```suggestion
    transition: background-color 0.3s ease-in-out, transform 0.3s ease-in-out;
```

asakevych (changes requested):
Good job! just some small changes left 😉

choeqq (changes requested):
Waiting for any updates @Vira-V 👀

Vira-V:
Updated added

asakevych (approved):
Great job!! 🚀 🔥 🔥

PR #2424, opened 2023-07-09
task v1.1 fixed

Miponkin:
- [DEMO LINK](https://Miponkin.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,29 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  padding: $paddingCal;
+  gap: $gap;
+  max-width: $calendarSize;
+
+  @each $num, $day in $days {
+    &--start-day--#{$day} :first-child {
+      margin-left: $num * ($cellSize + $gap);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


volodymyr-soltys97 (changes requested):
This part of the task is incomplete
image
image

Miponkin:
FIXED

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,33 @@
...
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
```
agniya-i, line 11: It would be better to move these styles to separate file or `main.scss`, since them are not related to the `calendar` block

src/styles/blocks/cell.scss
```diff
@@ -0,0 +1,27 @@
...
+  height: $cellSize;
+  width: $cellSize;
+  align-items: center;
+  justify-content: center;
+  background-color: $color;
+
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+
+  transition: 0.5s;
```
agniya-i, line 14: Don't forget to specify the properties to which a transition effect should be applied
```suggestion
  transition: transform 0.5s, ...;
```

agniya-i (approved):

PR #2425, opened 2023-07-09
Develop

kuzgan:
[DEMO LINK](https://kuzgan.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <section class="calendar calendar--start-day-sun calendar--month-length-28">
+    <div class="calendar calendar--start-day-wed calendar--month-length-28">
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
@@ -39,6 +39,6 @@
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
-    </section>
+    </div>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <section class="calendar calendar--start-day-sun calendar--month-length-28">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <section class="calendar calendar--start-day-sun calendar--month-length-28">
```
mvjl000: I like the fact that you want to bring semantics to your html but in this particular case, section doesn't make much sense. I would prefer to see div here

[section tag use cases](https://www.w3schools.com/tags/tag_section.asp)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
+@import "./utils/_variables" ;
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  width: 706px;
```
mvjl000: This is like a magic number, it is not clear why is it set to 706px and if you change for instance dimensions of calendar-day things will break up.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,11 @@
...
+$font-family: arial, sans-serif;
+$font-size: 30px;
+$days:
+  "mon" 0px,
+  "tue" 101px,
+  "wed" 202px,
+  "thu" 303px,
+  "fri" 404px,
+  "sat" 505px,
+  "sun" 606px;
```
mvjl000: Hard-coded values again. Changes pixels to indexes and then you will be able to calculate the margin basing on that ;)
```suggestion
$days:
  "mon" 0px,
  "tue" 1,
  "wed" 2,
  "thu" 3,
etc...
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: $day-size;
+    width: $day-size;
+    border: 1px solid #000;
+    background: #eee;
+    padding: 10px;
+    transition: all ease 0.5s;
```
mvjl000: Don't use `all`. Instead, list all values for this transition. It will make it way easier to understand what is it for.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+      cursor: pointer;
+      transform: translateY(-20px);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i}) {
+        &::before {
+          content: "#{$i}";
+          font-size: $font-size;
+          font-family: $font-family;
```
mvjl000: Things that never change like font-size and font family should be outside the loop ;)

mvjl000 (changes requested):

mvjl000 (approved):
🚀

PR #2426, opened 2023-07-09
add task solution

vensecue:
- [DEMO LINK](https://vensecue.github.io/layout_calendar/)
- [TEST REPORT LINK](https://vensecue.github.io/layout_calendar/report/html_report/)


src/index.html
```diff
@@ -7,6 +7,36 @@
...
+      <div class="calendar__day">20</div>
+      <div class="calendar__day">21</div>
+      <div class="calendar__day">22</div>
+      <div class="calendar__day">23</div>
+      <div class="calendar__day">24</div>
+      <div class="calendar__day">25</div>
+      <div class="calendar__day">26</div>
+      <div class="calendar__day">27</div>
+      <div class="calendar__day">28</div>
+      <div class="calendar__day">29</div>
```
mvjl000: You musn't set the content in that way. It has to be done by a ::before and a @for loop

src/index.html
```diff
@@ -7,6 +7,36 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
```
mvjl000: You are missing the key classes that allow changing the number of days and starting day. Take a look at the checklist please.

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+  min-height: 100vh;
+  display: flex;
+  align-items: center;
+}
+
+.calendar {
+  display: flex;
+  margin: 0 auto;
+  flex-wrap: wrap;
+  width: calc(7 * (#{$cell-size} + 2 * #{$gap} + 2 * #{$padding}));
```
mvjl000: I think your calculations need some adjustments, let's simplify it:

7 cells + 6 gaps

No need for paddings

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
 body {
   margin: 0;
+  min-height: 100vh;
+  display: flex;
+  align-items: center;
+}
+
+.calendar {
+  display: flex;
+  margin: 0 auto;
```
mvjl000: You are already centering the calendar vertically with body flex. Let's get rid of this margin and center it with body (justify-content)

src/styles/mixins.scss
```diff
@@ -0,0 +1,12 @@
...
+  height: $cell-size;
+  border: 1px solid $border-color;
+  background-color: $background-color;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+  cursor: pointer;
+  transition: transform 0.5s;
+}
```
mvjl000: There is no need for such a mixin, put all __day properties in a standard way.

If you want to use a mixin, let's do it in more reasonable case:
```suggestion
@mixin dimensions {
  width: $cell-size;
  height: $cell-size;
}
```

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+        .calendar--month-length-30 &:nth-child(n + 31) {
+          display: none;
+        }
+      }
+    }
+  }
+}
+
+.calendar__day:first-child {
+  margin-left: calc(7 * #{$cell-size} + #{$padding} - 3 * #{$gap});
```
mvjl000: The way you should approach this is with `@each`

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+  flex-wrap: wrap;
+  width: calc(7 * (#{$cell-size} + 2 * #{$gap} + 2 * #{$padding}));
+  padding: $padding;
+  gap: $gap;
+
+  &__day {
+    @include cell;
+
+    &::before {
+      content: "";
```
mvjl000: content should be set with the help of @for loop

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+    }
+
+    &:hover {
+      background-color: pink;
+      transform: translateY(-10px);
+    }
+  }
+  .calendar__day {
```
mvjl000: You are already in a calendar block, no need for this

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i}) {
+        @extend .calendar__day;
+        .calendar--month-length-28 &:nth-child(n + 29),
+        .calendar--month-length-29 &:nth-child(n + 30),
+        .calendar--month-length-30 &:nth-child(n + 31) {
+          display: none;
+        }
+      }
+    }
```
mvjl000: You need to completely rebuild this loop. Some hints:

- The range should be different. In this case you care only about 28-31
- You need to point to calendar with particular day modificator and then inside of it point to calendar__day with nth-child pseudo class

mvjl000 (changes requested):

choeqq (changes requested):
Waiting for any updates @vensecue 👀

choeqq (approved):
Nicely done 🚀

PR #2427, opened 2023-07-10
Solution

Patryk-Filipiak:
- [DEMO LINK](https://Erchu420.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Erchu420.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -15,12 +15,11 @@
     justify-content: center;
     border: 1px solid black;
     background-color: #eee;
-    transition: transform 0.5s;
-    counter-increment: day;
+    transition: all 0.5s;
+    cursor: pointer;
 
     &:hover {
       background-color: pink;
-      cursor: pointer;
       transform: translateY(-20px);
     }
 
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -33,13 +33,13 @@
   }
 
   @each $i, $day in $startDay {
-    &--start-day--#{$day} .calendar__day:first-child {
+    &--start-day_#{$day} .calendar__day:first-child {
       margin-left: $i * $day-size + $calendar-gap * $i;
     }
   }
 
   @for $i from 28 through 30 {
-    &--month-length--#{$i} {
+    &--month-length_#{$i} {
       .calendar__day:nth-child(n + #{$i + 1}) {
         display: none;
       }
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,49 @@
+.calendar {
+  position: relative;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-gap;
+  width: calc(#{$days-count} * #{$day-size} + #{$gapCount} * #{$calendar-gap});
+  margin: 20px auto;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+
+  &__day {
+    width: $day-size;
+    height: $day-size;
+    box-sizing: border-box;
+    display: flex;
+    align-items: center;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+    counter-increment: day;
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+    }
+
+    &::before {
+      content: counter(day);
```
mvjl000: You should set content using additional @for loop - don't use css counter

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
+.calendar {
+  position: relative;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-gap;
+  width: calc(#{$days-count} * #{$day-size} + #{$gapCount} * #{$calendar-gap});
+  margin: 20px auto;
```
mvjl000: The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
+.calendar {
+  position: relative;
```
mvjl000: You don't need position relative here

mvjl000 (changes requested):

src/styles/blocks/calendar.scss
Patryk-Filipiak: removed in 'Fixed Solution'

src/styles/blocks/calendar.scss
Patryk-Filipiak: changed to flex in 'Fixed Solution'

src/styles/blocks/calendar.scss
Patryk-Filipiak: changed in 'Fixed Solution'

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+  &__day {
+    width: $day-size;
+    height: $day-size;
+    box-sizing: border-box;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border: 1px solid black;
+    background-color: #eee;
+    transition: transform 0.5s;
```
mvjl000: Add the backround property to your transition because right now you are only animating the transform.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    counter-increment: day;
```
mvjl000: You are not using css counters anymore here so get rid of all counter related properties ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
```
mvjl000: I would move `cursor: pointer;` up, into __day.

mvjl000 (changes requested):
Minor changes needed. Please correct it and your solution will be on point 👌

src/styles/blocks/calendar.scss
Patryk-Filipiak: changed to 'all' in 'Solution 2'

src/styles/blocks/calendar.scss
Patryk-Filipiak: i was forgot about this, removed in 'Solution 2'

src/styles/blocks/calendar.scss
Patryk-Filipiak: moved in 'Solution 2'

choeqq (approved):
All good! Nicely done ✅

PR #2430, opened 2023-07-10
Calendar

ghost:
- [DEMO LINK](https://WaveOfDandelions.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
   margin: 0;
 }
+
+$grey: #eee;
+$square-side: 100px;
+$calendar-gap: 1px;
+$calendar-day-border: 1px;
+$calendar-day-space: $square-side + $calendar-gap;
+$calendar-padding: 10px;
+$calendar-width: ($square-side + $calendar-gap) * 7;
```
mvjl000: Create a separate file for variables ;)

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+
+.calendar {
+  width: $calendar-width;
+  margin: 0 auto;
```
mvjl000: The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: $calendar-gap;
+  padding: $calendar-padding;
+  padding-top: $calendar-padding * 3;
```
mvjl000: I don't think this two variables make sense

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+    justify-content: center;
+    align-items: center;
+    transition: 0.5s;
+  }
+
+  @for $index from 1 through 31 {
+    &__day:nth-child(#{$index})::before {
+      content: "#{$index}";
+      font-family: Arial, sans-serif;
+      font-size: 30px;
```
mvjl000: This should be inside of day. Inside of loops, put only the properties that are relying on loop - in this case content.

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+    }
+  }
+
+  @for $index from 1 through 31 {
+    &__day:nth-child(#{$index}):hover {
+      background-color: pink;
+      cursor: pointer;
```
mvjl000: I would like you to move `cursor: pointer` up, so it's located inside of __day

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+  &__day {
+    box-sizing: border-box;
+    background-color: $grey;
+    width: $square-side;
+    height: $square-side;
+    border: $calendar-day-border solid black;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: 0.5s;
```
mvjl000: Make sure to list all styles that you apply transition to - don't use `all` either.

mvjl000 (changes requested):

src/styles/main.scss
WaveOfDandelions: > Create a separate file for variables ;)

Done

src/styles/main.scss
WaveOfDandelions: > The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body

Done, I was set 100vh for body, and use flex alignment

src/styles/main.scss
WaveOfDandelions: > I would like you to move `cursor: pointer` up, so it's located inside of __day

Done

src/styles/main.scss
WaveOfDandelions: > Make sure to list all styles that you apply transition to - don't use `all` either.

Ready!

src/styles/main.scss
WaveOfDandelions: > This should be inside of day. Inside of loops, put only the properties that are relying on loop - in this case content.

Changed

src/styles/main.scss
WaveOfDandelions: > I don't think this two variables make sense

I was apply one padding value

choeqq (approved):
All changes applied - nice! ✅

PR #2431, opened 2023-07-10
add task solution

griveous199:
[DEMO LINK](https://Nightycake.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body class="main">
-    <div class="calendar calendar--month-length calendar--start-day">
+    <div class="calendar calendar--month-length-29 calendar--start-day-1">
       <div class="calendar__field"></div>
       <div class="calendar__field"></div>
       <div class="calendar__field"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="main">
+    <div class="calendar calendar--month-length calendar--start-day">
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
+      <div class="calendar__field"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="main">
+    <div class="calendar calendar--month-length calendar--start-day">
```
mvjl000: You are not using the modificator. Take a look at the check list. You should be able to change the number of displayed days with help of `calendar--month-length-#` i.e. `calendar--month-length-29.

Same thing with start day. You should set proper margin-left basing on the value passed there.

src/styles/main.scss
```diff
@@ -1,3 +1,56 @@
...
+  &__field {
+    background-color: #eee;
+    height: 100px;
+    width: 100px;
+    border: solid black 1px;
+    font-size: 30px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 0.5s;
```
mvjl000: Instead of `all` list all the properties that you want to animate.

mvjl000: It's more readable and you know exactly what is affected by it.

src/index.html
mvjl000: Okay, I can see that it's working after I add this directly into html. But you can add some default value to your html

src/styles/main.scss
```diff
@@ -1,3 +1,56 @@
...
+  height: 100vh;
+  font-family: Arial, sans-serif;
+  margin: 9px;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  margin: auto;
+  width: 721px;
```
mvjl000: Avoid such a hard-coded values.
This is like a magic number, it is not clear why is it set to 721px and if you change for instance dimensions of calendar-day things will break up.

src/styles/main.scss
```diff
@@ -1,3 +1,56 @@
...
+  padding: 10px;
+  gap: 1px;
+
+  &__field {
```
mvjl000, line 21: Use variables for field dimensions (width, height) and also for calendar gap. With this, you will be able to calculate everything else

src/styles/main.scss
```diff
@@ -1,3 +1,56 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 0.5s;
+  }
+
+  &__field:hover {
+    transform: translate(0, -20px);
+    background-color: pink;
+    cursor: pointer;
```
mvjl000: I would prefer `cursor: pointer` to be outside the hover block, inside field.

mvjl000 (changes requested):

choeqq (changes requested):
Waiting for any updates @Nightycake 👀

src/styles/variables.scss
```diff
@@ -0,0 +1,3 @@
+$gaps: 1px;
+$default-font: 30px;
+$calendar-width: calc(103px * 7);
```
choeqq: you should also take gap and padding into your calendar witdth calculation

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+  }
+
+  @for $month-lenght from 28 through 31 {
+    &--month-length--#{$month-lenght}
+    .calendar__field:nth-child(n + #{$month-lenght + 1}) {
+      display: none;
+    }
+  }
+
+  $start-day: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
```
choeqq: please move those to variables file and you're good to go 🚀

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  margin: auto;
+  width: $calendar-width;
+  gap: $gaps;
+
+  &__field {
+    background-color: #eee;
+    height: 100px;
+    width: 100px;
```
choeqq: please use `field_size` variable for cell size

choeqq (changes requested):
Almost there - small changes needed 🚀

choeqq (approved):
Nicely done 🚀

PR #2435, opened 2023-07-10
Develop

jakubMikolajczykTech:
[DEMO LINK](https://kabaktech.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
-body {
-  margin: 0;
+.calendar {
+  width: 707px;
+  margin: 0 auto;
```
mvjl000, line 3: The calendar should be centered both horizontally and vertically. Maybe try to add display flex to parent, which is body.

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
-body {
-  margin: 0;
+.calendar {
```
mvjl000, line 1: `.calendar {}` should not be nested inside of body. Please change it

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+  width: 707px;
```
mvjl000, line 2: Avoid such a hard-coded values.
This is like a magic number, it is not clear why is it set to 707px and if you change for instance dimensions of calendar-day things will break up.

Please create a separate file for variables

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+  margin: 0 auto;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 30px 10px 10px;
```
mvjl000, line 7: You won't need this paddings after centering calendar inside of body.

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+
+  $days: mon, tue, wed, thu, fri, sat, sun;
```
mvjl000, line 9: This should also go to variables file I mentioned before

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+      font-family: Arial, sans-serif;
+      font-size: 30px;
+    }
+  }
+
+  &__day {
+    box-sizing: border-box;
+    background-color: #eee;
+    width: 100px;
+    height: 100px;
```
mvjl000, line 39: You need to create variables for day dimensions. Then you will be able to calculate evertything basing on them.

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+    border: 1px solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: transform 0.5s;
+
+    &:first-child:hover,
+    &:nth-child(n +2):hover {
```
mvjl000, line 47: I don't get those selectors, it's overcomplicated. Instead just do:
```suggestion
&:hover {
```

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+      background-color: pink;
+      cursor: pointer;
```
mvjl000, line 49: I would move `cursor: pointer;` up so it's inside of __day

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+  &__day {
+    box-sizing: border-box;
+    background-color: #eee;
+    width: 100px;
+    height: 100px;
+    border: 1px solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: transform 0.5s;
```
mvjl000, line 44: List all properties for transition (but don't use `all` either). Background-color is not listed and thus, it's not animated - it turns pink immediately

mvjl000 (changes requested):

choeqq (changes requested):
Waiting for any updates @kabakTech 👀

PR #2436, opened 2023-07-10
task done

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+
```
mvjl000: The calednar should be centered both horizontally and vertically. Adding flex to body will solve this problem ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.calendar {
+  display: flex;
+  width: #{$calendar-width};
+  flex-wrap: wrap;
```
mvjl000: Your calendar looks very tightly-packed. Please add gap of 1px.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$calendar-gap: 1px;
+$border-field: 1px;
+$font-size: 30px;
+$field-width: 100px;
+$space: calc(#{$field-width} + #{$calendar-gap});
+$calendar-width: calc(#{$space} * 7);
```
mvjl000: You need to change those calculations a bit. When having 7 columns in a calendar, you should have 6 gaps

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,13 @@
+@each $number, $day in $name_days {
+  .calendar--start-day-#{$day} :first-child {
+    margin-left: calc(#{$number} * #{$field-width});
+  }
+}
+
+@for $i from 28 through 31 {
+  .calendar--month-lenght-#{$i} {
+    &:nth-child(n + #{$i + 1}) {
```
mvjl000: Changing the number of days in your solution doesn't work as it should. I think this is the right place you should change in order to fix it. Maybe point to calendar__field?

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+    justify-content: center;
+    align-items: center;
+    transition: all 0.3s;
+
+    @for $index from 1 through 31 {
+      &:nth-child(#{$index}) {
+        &::before {
+          content: "#{$index}";
+          font-family: Arial, Helvetica, sans-serif;
+          font-size: #{$font-size};
```
mvjl000: Inside of loops, put only the styles that are changing, in this case only content should be inside of loop. Move the font properties up.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+
+  &__fields {
+    @include field-size($field-width);
+    gap: #{$calendar-gap};
+    background-color: #eee;
+    border: #{$border-field} solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 0.3s;
```
mvjl000: List all the properties that you are applying transition to (transform and background-color). It will be way easier to see what this transition is affecting ;)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+        &::before {
+          content: "#{$index}";
+          font-family: Arial, Helvetica, sans-serif;
+          font-size: #{$font-size};
+        }
+
+        &:hover {
+          cursor: pointer;
+          background-color: pink;
+          transform: translateY(-5px);
```
mvjl000: Two things here:

- I would move the `cursor: pointer;` up. So it's not in hover but inside of __field,
- According to the checklist, you should move cell up by 20px ;)

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sun calendar--month-lenght-28">
+      <div class="calendar__fields"></div>
```
mvjl000: Honestly I would change this class name to a signular form

mvjl000 (changes requested):

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,13 @@
+@each $number, $day in $name_days {
+  .calendar--start-day-#{$day} :first-child {
+    margin-left: calc(#{$number} * #{$field-width});
+  }
+}
+
+@for $i from 28 through 31 {
+  .calendar--month-length-#{$i} {
+    .calendar__fields:nth-child(#{$i}) {
```
JoyCoffeeAddict: It was not working on q&a, because right now you are accessing only one day in from calendar, that is equal to moth length
e.g. .calendar--month-length-30 will only target 30-th child 

Try to come up with solution to select all children after specified day
<details>
  <summary>TIP</summary>
  
use nth-child() with n
  ```
</details>

JoyCoffeeAddict (changes requested):

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$calendar-gap: 1px;
+$border-field: 1px;
+$font-size: 30px;
+$field-width: 100px;
+$space: calc(#{$field-width} + #{$calendar-gap});
+$calendar-width: calc((#{$space} + #{$border-field}) * 7 );
```
asakevych: it would be easier to calculate if you set `box-siizing: border-box` for each element using `*` selector, then you don't need to calculate borders, it would be enough - `$field-width * 7 + $calendar-gap * 6` and variable space would be unnecessary

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin field-size ($size) {
+  width: $size;
+  height: $size;
+}
```
asakevych: i thing you don't really need this mixin, it is used just once 😉

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,13 @@
+@each $number, $day in $name_days {
+  .calendar--start-day-#{$day} :first-child {
+    margin-left: calc(#{$number} * #{$space});
```
asakevych, line 3: 1. I think here you just need `($field-width + $calendar-gap) * $ number`
2. it's not necessary to write calc using scss 😉

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    @include field-size($field-width);
+    background-color: #eee;
+    border: $border-field solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-duration: 0.5s;
+    transition-delay: 0.5s;
+    transition-timing-function: ease-in-out;
+    font-family: Arial, Helvetica, sans-serif;
```
asakevych: it's a good practice to define font-family for the whole body

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  &__field {
+    @include field-size($field-width);
+    background-color: #eee;
+    border: $border-field solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-duration: 0.5s;
+    transition-delay: 0.5s;
+    transition-timing-function: ease-in-out;
```
asakevych: please define also transition-property with each property that you need, not just 'all' - https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

asakevych (changes requested):
Good job, but still some changes are needed 😉

src/styles/utils/_functions.scss
patryk177m: $space: calc(#{$field-width} + #{$calendar-gap});
without calc it won't work

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+    height: $field-width;
+    background-color: #eee;
+    border: $border-field solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-duration: 0.5s;
+    transition-delay: 0.5s;
+    transition-timing-function: ease-in-out;
+    transition-property: all;
```
choeqq: please say explicitly which CSS properties you want to transition, because now even changing calendar length is weirdly animated

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+      &:nth-child(#{$index}) {
+        &::before {
+          content: "#{$index}";
+        }
+
+        &:hover {
+          cursor: pointer;
+          background-color: pink;
+          transform: translateY(-20px);
+        }
```
choeqq: those styles are just hover, they don't have to be inside the loop

choeqq (changes requested):
Looks good, little change needed and you're good to go 🚀

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_calendar/)

JoyCoffeeAddict (approved):
Well done

PR #2442, opened 2023-07-10
add task solution

PawelP2001:


src/styles/main.scss
choeqq, line 1: please follow [checklist](https://github.com/mate-academy/layout_calendar/blob/master/checklist.md) point 3 file structure

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  "sun": 0,
+  "mon": 1,
+  "tue": 2,
+  "wed": 3,
+  "thu": 4,
+  "fri": 5,
+  "sat": 6,
+);
+
+$monthLengths: 28, 29, 30, 31;
```
choeqq: please use separate file for variables

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  gap: $gap;
+
+  &__day {
+    box-sizing: border-box;
+    width: $cellSize;
+    height: $cellSize;
+    background-color: #eee;
+    position: relative;
+    border: 1px solid black;
+    transition: all ease-in-out 0.3s;
```
choeqq: please don't set transition to all - list all properties you want to animate

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  "tue": 2,
+  "wed": 3,
+  "thu": 4,
+  "fri": 5,
+  "sat": 6,
+);
+
+$monthLengths: 28, 29, 30, 31;
+
+.calendar {
```
choeqq: please center calendar both vertically and horizontally

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-start;
+  max-width: $calendarWidth;
+  margin: 0 auto;
+  padding: $padding;
+  gap: $gap;
+
+  &__day {
```
choeqq: we can also use flex to ceneter `__day` - then there will be no point in styling `::before`

choeqq (changes requested):
Almost there, small changes needed 🚀

src/styles/main.scss
choeqq, line 1: please center calendar

and try to make your calendar work with custom values like 24 - [checklist ](https://github.com/mate-academy/layout_calendar/blob/master/checklist.md)point 10

choeqq (changes requested):
little changes needed 🚀

src/styles/main.scss
PawelP2001: @choeqq i've just now centered it, but i dont necessarily understand the part of it being able to work with custom values, any clues how to do that?

src/styles/main.scss
```diff
@@ -1,3 +1,2 @@
-body {
-  margin: 0;
-}
```
choeqq: Please give body those styles: 

```
display: flex;
justify-content: center;
align-items: center;
min-height: 100vh;
```

choeqq (changes requested):
One little change needed 😄 

When it comes to working with custom days value - your calendar should handle custom month length like 23 correctly - it means that there should be 23 days displayed,now when I provide some custom value other than 28-31 every cell's content vanishes

choeqq (approved):
Looks good! Approved ✅

PR #2443, opened 2023-07-11
add task solution

ndovgyi:
[DEMO LINK] https://ndovgyi.github.io/layout_calendar/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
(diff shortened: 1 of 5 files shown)


VitaliyBondarenko1982 (changes requested):
Calendar doesn't work
after change modifiers calendar should start from start-day in modifier and should be visible month-length days which set in modifier.
Screenshot from 2023-07-11 16-49-04


and pay attention, please on task description
Screenshot from 2023-07-11 16-46-02

mykhalenych (approved):

PR #2446, opened 2023-07-11
add task solution

szymonpachucki:
[DEMO LINK](https://szymonpachucki.github.io/layout_calendar/)
- [TEST REPORT LINK](https://szymonpachucki.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,33 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $border-size;
+  max-width: $calendar-width;
+
+  &__day {
+    display: flex;
+    width: $cell-size;
+    height: $cell-size;
+    border: $border-size solid $black;
+    background-color: $gray;
+    font-size: $font-size;
+    text-align: center;
+    position: relative;
+    cursor: pointer;
+
+    &::before {
+      content: "";
+      font-size: 30px;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+    }
+
+    &:hover {
+      background-color: $pink;
+      transition: transform 0.5s;
+      transform: translateY(-20px);
+    }
+  }
+}
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -25,9 +25,9 @@
     }
 
     &:hover {
-      background-color: $pink;
-      transition: transform 0.5s;
+      transition: transform 0.5s, background-color 0.5s;
       transform: translateY(-20px);
+      background-color: $pink;
     }
   }
 }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,34 +1,16 @@
-// Colors
-$gray: #eee;
-$pink: #ff69b4;
-$black: #000;
+@import "variables";
+@import "calendar";
 
-// Sizes
-$cell-size: 100px;
-$border-size: 1px;
-$font-size: 30px;
-$padding-size: 10px;
-
-// Calculation variables
-$calendar-width: (7 * $cell-size) + (8 * $border-size) + (2 * $padding-size);
-$column-count: 7;
-
-// Mixins
-@mixin hover-styles {
-  cursor: pointer;
-  background-color: $pink;
+body {
+  display: flex;
+  justify-content: center;
 }
 
-// Modifier classes
-$start-days: (
-  "mon" 0,
-  "tue" 1 * ($cell-size + ($border-size * 2)) + 1px,
-  "wed" 2 * ($cell-size + $border-size * 2) + 2px,
-  "thu" 3 * ($cell-size + $border-size) + 6px,
-  "fri" 4 * ($cell-size + $border-size) + 8px,
-  "sat" 5 * ($cell-size + $border-size * 2) + 5px,
-  "sun" 6 * ($cell-size + $border-size * 2) + 5px
-);
+* {
+  font-family: Arial, Helvetica, sans-serif;
+  box-sizing: border-box;
+}
+
 @each $day, $margin in $start-days {
   .calendar--start-day--#{$day} {
     .calendar__day:first-child {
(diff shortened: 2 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+// Sizes
+$cell-size: 100px;
+$border-size: 1px;
+$font-size: 30px;
+$padding-size: 10px;
+
+// Calculation variables
+$calendar-width: (7 * $cell-size) + (8 * $border-size) + (2 * $padding-size);
+$column-count: 7;
+
```
asakevych: please move all variables to separate file

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+// Mixins
+@mixin hover-styles {
+  cursor: pointer;
+  background-color: $pink;
+}
+
```
asakevych: this mixin in unnecessary, those styles can be used just once 😉

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+// Modifier classes
+$start-days: (
+  "mon" 0,
+  "tue" 1 * ($cell-size + ($border-size * 2)) + 1px,
+  "wed" 2 * ($cell-size + $border-size * 2) + 2px,
+  "thu" 3 * ($cell-size + $border-size) + 6px,
+  "fri" 4 * ($cell-size + $border-size) + 8px,
+  "sat" 5 * ($cell-size + $border-size * 2) + 5px,
+  "sun" 6 * ($cell-size + $border-size * 2) + 5px
+);
```
asakevych: 1. move it to variables file
2. you shouldn't have calculation like this here, you can just write @each loop and use siple values like: 
```suggestion
$start-days: (
  "mon" 0,
  "tue" 1,
  "wed" 2,
  "thu" 3,
  "fri" 4,
  "sat" 5,
  "sun" 6
);
```

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+}
+
+// Modifier classes for month length
+@for $i from 28 through 31 {
+  .calendar--month-length--#{$i} {
+    .calendar__day:nth-child(n+#{$i + 1}) {
+      display: none;
+    }
+  }
+}
```
asakevych, line 29: it's unnecessary, try to think about other solution, using `$start-days` 😉

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+
+// Styles
+.calendar {
```
asakevych: every block should be in separate file - create block/calendar.scss for this block

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+  margin: auto;
```
asakevych: you can position it using flex on parent, which is body here

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: $border-size;
+  padding: $padding-size;
```
asakevych: this padding is unnecessary, you can remove it and simplify calculating width for this container 😉

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+$black: #000;
+
+// Sizes
+$cell-size: 100px;
+$border-size: 1px;
+$font-size: 30px;
+$padding-size: 10px;
+
+// Calculation variables
+$calendar-width: (7 * $cell-size) + (8 * $border-size) + (2 * $padding-size);
```
asakevych: 1. you shouldn't calculate border if you set `box-sizing: border-box` for all elements, using `*`
2.  padding is unnecessary in your container, you can remove it and simplify your calculation 😉 
3.  remember about adding gaps here too, you can also create variable - example 
`$gap: 1px` - use for `.calendar` and to calculate here `6*$gap`

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+    text-align: center;
+    position: relative;
+
+    &::before {
+      content: "";
+      font-size: 30px;
+      font-family: Arial, Helvetica, sans-serif;
+      position: absolute;
+      top: 50%;
+      left: 50%;
```
asakevych: 1. you don't need to use position absolute here, try to position it using display flex for `&__day`
2. it's a good practice to set font-family for the whole body

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+      transform: translate(-50%, -50%);
+    }
+
+    &:hover {
+      @include hover-styles;
```
asakevych: 1. it's better just to write styles here, mixin is unnecessary, it is used just once
2. add transform to make cell pop-up, i think you can implement it -> 

If you already know Transition and Transform
- Move the hovered cell up by 20px (use transform)
- All changes should be animated with the duration of 0.5s

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+  gap: $border-size;
+  padding: $padding-size;
+  max-width: $calendar-width;
+
+  &__day {
+    width: $cell-size;
+    height: $cell-size;
+    border: $border-size solid $black;
+    background-color: $gray;
+    font-family: Arial, Helvetica, sans-serif;
```
asakevych: i think you can define font-family for the whole body and delete it from every element 😉

src/styles/main.scss
```diff
@@ -1,3 +1,88 @@
...
+    font-size: $font-size;
+    text-align: center;
+    position: relative;
```
asakevych: 1. it's unnecessary here, try to use flex to position `:before`
2. add  `cursor: pointer` here, not to hover

asakevych (changes requested):
Good job! just few changes needed 😉

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

src/styles/calendar.scss
```diff
@@ -0,0 +1,33 @@
...
+      font-size: 30px;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+    }
+
+    &:hover {
+      background-color: $pink;
+      transition: transform 0.5s;
```
choeqq: you also need to place background-color in there, now it's changing instantly 
please move transition property to &__day, on hover it makes cells behave little odd

choeqq (changes requested):
One little change and you're good to go 🚀

choeqq (approved):
Nice - approved ✅

PR #2460, opened 2023-07-14
calendar was created

maksym-mishchanchuk:
- [DEMO LINK](https://maksym-mishchanchuk.github.io/layout_calendar/)


src/styles/utils/variables.css
```diff
@@ -0,0 +1,3 @@
+
+
+/*# sourceMappingURL=variables.css.map */
```
roman-mirzoian: This file seems to be empty. It should define some styles or variables based on the task description.

src/styles/blocks/calendar.css
```diff
@@ -0,0 +1,17 @@
+/* Error: Undefined variable.
```
roman-mirzoian: There's an error in your SCSS file, it should be fixed.

src/styles/blocks/calendar.css
```diff
@@ -0,0 +1,17 @@
...
+
+body::before {
+  font-family: "Source Code Pro", "SF Mono", Monaco, Inconsolata, "Fira Mono",
+      "Droid Sans Mono", monospace, monospace;
+  white-space: pre;
+  display: block;
+  padding: 1em;
+  margin-bottom: 1em;
+  border-bottom: 2px solid black;
+  content: "Error: Undefined variable.\a   \2577 \a 4 \2502    width: (7 * $size-day) + 2 * $padding-month;\a   \2502                ^^^^^^^^^\a   \2575 \a   calendar.scss 4:15  root stylesheet";
```
roman-mirzoian: The content property is displaying an error message, which is unneeded in the final CSS file. Remove this after fixing the SCSS error.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,4 @@
+$size-day: 100px;
```
roman-mirzoian: For variable names, it is better to give more precise names, for example: `$cell-size`, `$day-cell-size`.
This applies to all variables.

src/styles/main.css
```diff
@@ -0,0 +1,188 @@
...
+.calendar--month-length--29 .day:nth-child(n+30) {
+  display: none;
+}
+.calendar--month-length--30 .day:nth-child(n+31) {
+  display: none;
+}
+.calendar--month-length--31 .day:nth-child(n+32) {
+  display: none;
+}
+.calendar__start-day--mon > :first-child {
```
roman-mirzoian, line 26: The start day modifier does not account for the 1px gap between days. Each day should be offset by 101px, not 100px.

src/styles/main.css
```diff
@@ -0,0 +1,188 @@
...
+.calendar__start-day--sun > :first-child {
+  margin-left: 606px;
+}
+
+.day {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 100px;
+  height: 100px;
```
roman-mirzoian, line 53: The grey color for the square is not specified. It is better to use a specific grey color code, for example: #808080.

src/styles/main.css
```diff
@@ -0,0 +1,188 @@
+body {
+  margin: 0;
+  font-family: Arial, Baghdad, sans-serif;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: 720px;
```
roman-mirzoian: The width of the calendar is larger than required by the task. If each day is represented by a 100px square and there are 7 columns with a 1px gap, the width should be 707px (7*100 + 6*1).

src/styles/main.css
```diff
@@ -0,0 +1,188 @@
...
+  margin-left: 404px;
+}
+.calendar__start-day--sat > :first-child {
+  margin-left: 505px;
+}
+.calendar__start-day--sun > :first-child {
+  margin-left: 606px;
+}
+
+.day {
```
roman-mirzoian, line 48: The task specifies that the calendar should have 10px padding, which is not included in the current implementation.

src/styles/main.css
```diff
@@ -0,0 +1,188 @@
...
+  box-sizing: border-box;
+  border: 1px solid black;
+  transition: 0.3s;
+}
+.day:hover {
+  transform: translateY(-20px);
+  cursor: pointer;
+  background-color: mistyrose;
+}
+.day:nth-child(1)::before {
```
roman-mirzoian, line 65: The day numbers are hard-coded using pseudo-elements. This is not maintainable and should be generated dynamically in the HTML or JavaScript.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,21 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: (7 * $size-day) + 2 * $padding-month;
```
roman-mirzoian: Try not to use "magic values", but to put them in variables, because now it is not clear what 7 and 2 are and whether it can be changed in the future, or put the full result of the calculation in a separate variable.

.stylelintrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   extends: [
-    "stylelint-config-recommended-scss",
-    "@mate-academy/stylelint-config"
+    // "stylelint-config-recommended-scss",
+    // "@mate-academy/stylelint-config"
```
roman-mirzoian: Why did you decide to disable these rules?

roman-mirzoian (changes requested):
The work is done well and the application works correctly, but the code itself needs to be improved a little.

src/styles/main.css
maksym-mishchanchuk: Use flex with 1px gap and limit its width to exactly 7 columns + 10px paddings

.stylelintrc.js
maksym-mishchanchuk: i have a problem with lint:

src/styles/utils/variables.scss
 1:1  ✖  Unknown rule annotation-no-unknown                               annotation-no-unknown                  
 1:1  ✖  Unknown rule function-no-unknown. Did you mean unit-no-unknown?  function-no-unknown                    
 1:1  ✖  Unknown rule custom-property-no-missing-var-function             custom-property-no-missing-var-function
 1:1  ✖  Unknown rule keyframe-block-no-duplicate-selectors               keyframe-block-no-duplicate-selectors  
 1:1  ✖  Unknown rule selector-anb-no-unmatchable

src/styles/blocks/day.scss
```diff
@@ -0,0 +1,26 @@
...
+.day {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: $cell-size;
+  height: $cell-size;
+  background-color: #eee;
+  box-sizing: border-box;
+  border: 1px solid black;
+  transition: 0.3s;
```
alexander-ignatow, line 12: Per task
> All changes should be animated with the duration of 0.5s

src/styles/blocks/day.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  &:hover {
+    transform: translateY(-20px);
+    cursor: pointer;
```
alexander-ignatow: Why are you changing the cursor style on element hover?

src/styles/blocks/day.scss
```diff
@@ -0,0 +1,26 @@
...
+    background-color: mistyrose;
+  }
+
+  @for $day from 1 through 31 {
+    &:nth-child(#{$day})::before {
+      content: "#{$day}";
+      font-size: 30px;
```
alexander-ignatow: Per checklist:

> [SASS] - Don't use SASS loops for styles that stay the same for all elements of the group, e.g. display or position.

src/styles/main.css
alexander-ignatow: Valid comment, but probably something is still wrong with your padding

src/styles/main.css
alexander-ignatow: Please fix this - paddings are still missing

image

alexander-ignatow (changes requested):
Good progress 👍  
Still some issues are needed to be fixed

src/styles/blocks/day.scss
BudnikOleksii, line 12: It's better to specify to what properties you need to apply the transition.
image

BudnikOleksii (approved):
👍

PR #2464, opened 2023-07-15
add task solution

Viktoriia-Vita:
- [DEMO LINK](https://viktoriia-vita.github.io/layout_calendar/)
- [TEST REPORT LINK](https://viktoriia-vita.github.io/layout_calendar/report/html_report/)

P.S. font-family Arial, Helvetica started with small letters because npm run lint

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--31">
+      <div class="calendar__day calendar__day--1 start-day-mon"></div>
+      <div class="calendar__day calendar__day--2"></div>
+      <div class="calendar__day calendar__day--3"></div>
+      <div class="calendar__day calendar__day--4"></div>
+      <div class="calendar__day calendar__day--5"></div>
+      <div class="calendar__day calendar__day--6"></div>
+      <div class="calendar__day calendar__day--7"></div>
+      <div class="calendar__day calendar__day--8"></div>
+      <div class="calendar__day calendar__day--9"></div>
+      <div class="calendar__day calendar__day--10"></div>
+      <div class="calendar__day calendar__day--11"></div>
+      <div class="calendar__day calendar__day--12"></div>
+      <div class="calendar__day calendar__day--13"></div>
+      <div class="calendar__day calendar__day--14"></div>
+      <div class="calendar__day calendar__day--15"></div>
+      <div class="calendar__day calendar__day--16"></div>
+      <div class="calendar__day calendar__day--17"></div>
+      <div class="calendar__day calendar__day--18"></div>
+      <div class="calendar__day calendar__day--19"></div>
+      <div class="calendar__day calendar__day--20"></div>
+      <div class="calendar__day calendar__day--21"></div>
+      <div class="calendar__day calendar__day--22"></div>
+      <div class="calendar__day calendar__day--23"></div>
+      <div class="calendar__day calendar__day--24"></div>
+      <div class="calendar__day calendar__day--25"></div>
+      <div class="calendar__day calendar__day--26"></div>
+      <div class="calendar__day calendar__day--27"></div>
+      <div class="calendar__day calendar__day--28"></div>
+      <div class="calendar__day calendar__day--29"></div>
+      <div class="calendar__day calendar__day--30"></div>
+      <div class="calendar__day calendar__day--31"></div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--31">
+      <div class="calendar__day calendar__day--1 start-day-mon"></div>
```
xiimiik: Each day div is missing the content for the day number. Consider adding the day number as content in each day div.

xiimiik: The class 'start-day-mon' suggests that the calendar will always start on Monday. Instead, consider making the start day dynamic based on the month and year.

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--31">
```
xiimiik: The class 'calendar--31' suggests that the calendar will always have 31 days. Instead, consider making the number of days in the calendar dynamic based on the month.

src/index.html
xiimiik: The days of the calendar are hard-coded in the HTML, which is not maintainable. Instead, consider generating the HTML for these days using JavaScript.

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+
+.calendar {
+  display: flex;
+  width: $column-width;
+  flex-wrap: wrap;
+  margin: 0 auto;
+  gap: $gap;
+
+  @for $month-length from 28 through 31 {
+    &--#{$month-length} :nth-child(#{$month-length}) {
```
xiimiik, line 19: The month-length modifier will not work as expected. It will hide all days after the specified day, but it will not hide the specified day itself. You should target from nth-child(n+1) instead of nth-child(n).

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+    height: $day-height;
+    background-color: $main-color;
+    box-sizing: border-box;
+    border: 1px solid black;
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+      transition-duration: 0.5s;
```
xiimiik: The transition property should be applied to the element, not on hover. If you apply it on hover, the transition will only work when you hover over the element, but not when you stop hovering.

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+    &--#{$month-length} :nth-child(#{$month-length}) {
+      & ~ .calendar__day {
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    display: flex;
+    flex-direction: column;
```
xiimiik: The flex-direction property is not needed here. The default value of flex-direction is row, which is what we want for the calendar cells.

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+        content: "#{$number}";
+        font-family: $main-font-family;
+        font-size: $main-font-size;
+        text-align: center;
+      }
+    }
+  }
+
+  @each $name, $index in $start-days {
+    .start-day-#{$name} {
```
xiimiik: The start-day modifier is being added to the wrong element. It should be added to the .calendar block, not to the .calendar__day element. Because of this, the shift by days of the week does not work correctly

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
   margin: 0;
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+  justify-content: center;
+}
+
+.calendar {
+  display: flex;
+  width: $column-width;
```
xiimiik, line 13: The width of the .calendar block should not be equal to $column-width, it should be calculated based on the number of columns, the width of the day cell, and the gaps between them. The correct calculation should look like this: width: $day-width * 7 + $gap * 6;

xiimiik (changes requested):
GJ, left few comments :)

src/index.html
Viktoriia-Vita: it's task for CSS advanced. we don't need use JS in this task

Viktoriia-Vita: How can I do this without JS???? I thought that would be enough added a class....

src/styles/main.scss
Viktoriia-Vita: sorry, I don`t understand.... if month have 28 days, so day 28th must be visible, but not 29th

Viktoriia-Vita:
Leave a comments and add changes on code

Esceype (approved):
Great job! But for the future don`t use the body tag selector to define the display flex and following properties, better make the container for set you block in center of page or find another way for get this.

PR #2477, opened 2023-07-18
add task solution

ydem1:
- [DEMO LINK](https://ydem1.github.io/layout_calendar/)
- [TEST REPORT LINK](https://ydem1.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/blocks/body.scss
+++ b/src/styles/blocks/body.scss
@@ -1,0 +1,5 @@
+.body {
+  &__calendar {
+    margin: 50px auto 0;
+  }
+}
--- a/src/styles/blocks/body.scss
+++ b/src/styles/blocks/body.scss
@@ -1,5 +1,5 @@
 .body {
-  &__calendar {
-    margin: 50px auto 0;
-  }
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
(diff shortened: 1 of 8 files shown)


src/styles/blocks/body.scss
```diff
@@ -0,0 +1,5 @@
+.body {
+  &__calendar {
+    margin: 50px auto 0;
```
andriy-shymkiv: I think you can do this direclty for `calendar` block

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+  &__body {
+    margin: 50px auto 0;
+  }
+
+  &__day {
+    position: relative;
+
+    width: $sizeDay;
+    height: $sizeDay;
+    padding: $paddingDay;
```
andriy-shymkiv: no need to set padding here, because you positioning text inside with `position: absolute`

andriy-shymkiv (changes requested):
good job! left comments

src/styles/blocks/body.scss
ydem1: Is it possible to set an external geometry (margin) for a block?

Esceype (approved):
Great job!!!

PR #2482, opened 2023-07-19
Layout Calendar add tasks

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/layout_calendar/)

### DIFF
--- a/src/styles/utils/variables.scss
+++ b/src/styles/utils/variables.scss
@@ -1,0 +1,12 @@
+$start-day:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
+
+$paddings: 10px;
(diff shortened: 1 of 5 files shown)


Anna-Tkachenko (changes requested):
Check functionality for `month-length` modifier, looks like it doesn't work
image

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,12 @@
+$start-day:
+  "mon" 0,
+  "tue" 1,
+  "wed" 2,
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
+
+$paddings: 10px;
```
roman-mirzoian, line 10: It is better to give a clearer name so that it is immediately clear what exactly these `paddings` are for.

roman-mirzoian (approved):
Well done 👍

PR #2485, opened 2023-07-19
Develop

AndrMar1939:
- [DEMO LINK](https://AndrMar1939.github.io/layout_calendar/)
- [TEST REPORT LINK](https://AndrMar1939.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,9 +1,4 @@
-$square-side: 100px;
-$calendar-padding: 10px;
-$calendar-gap: 1px;
-$calendar-width: ($square-side * 7)
-  + ($calendar-padding * 2) + ($calendar-gap * 6);
-$days-names: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+@import "./utils/variables.scss";
 
 body {
   margin: 0;
@@ -42,7 +37,7 @@ body {
     border: 1px solid #000;
     background-color: #eee;
 
-    transition: 0.5s;
+    transition: transform 0.5s;
 
     @for $i from 1 through 31 {
       &:nth-child(#{$i})::before {
@@ -65,7 +60,7 @@ body {
 
   @for $i from 28 through 31 {
     &--month-length--#{$i} > :not(:nth-child(-n + #{$i})) {
-      visibility: hidden;
+      display: none;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+
+  @each $day, $number in $days-names {
+    &--start-day--#{$day} > :first-child {
+      margin-left: calc((#{$square-side + $calendar-gap}) * #{$number});
+    }
+  }
+
+  @for $i from 28 through 31 {
+    &--month-length--#{$i} > :not(:nth-child(-n + #{$i})) {
+      visibility: hidden;
```
alexander-ignatow: `display: none` is a better choice here - please spend some time to find out why so.

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    width: $square-side;
+    height: $square-side;
+    border: 1px solid #000;
+    background-color: #eee;
+
+    transition: 0.5s;
```
> [STYLES] - Make sure to list all styles that you apply transition to.

src/styles/main.scss
alexander-ignatow, line 1: Per checklist:
> [BEM] - Create a separate file per each BEM block style that have the same name as the block

Also, it is better to have a dedicated file for variables and one file as main one.

alexander-ignatow (changes requested):
Clean solution - just few issues are needed to be addressed

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    width: $square-side;
+    height: $square-side;
+    border: 1px solid #000;
+    background-color: #eee;
+
+    transition: transform 0.5s;
```
BudnikOleksii: You can add a transition to a background-color too, it will be more smooth.

BudnikOleksii (approved):
Nicely done👍

src/styles/main.scss
alexander-ignatow: Mind @BudnikOleksii comment - check the syntax of `transition` - https://stackoverflow.com/questions/7048313/how-to-have-multiple-css-transitions-on-an-element

alexander-ignatow (changes requested):
Almost there!

Just a final minor issue left.

src/styles/main.scss
AndrMar1939: yep. Thank you for a useful link

Moroz-Dmytro (approved):
LGTM 👍

PR #2488, opened 2023-07-20
Develop

Oleh0293:
- [DEMO LINK](https://Oleh0293.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Oleh0293.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,68 @@
-body {
+$box-size: 100px;
+$spaces: 1px;
+$full-width: 720px;
+$box-color: #eee;
+$choosed-color: #fbbbcb;
+$days-of-weak:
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
-body {
+$box-size: 100px;
```
alexander-ignatow, line 1: `box-size` is a weird variable name as it sounds similar to CSS property `box-sizing`
try renaming it to `cell-size`, `day-size`, etc

It is still valid, but weird :)

alexander-ignatow (approved):
Clean solution 👍 

Please see my comment on variable naming, but still it LGTM

PR #2492, opened 2023-07-20
add calendar

ivannaromanovych:
[DEMO LINK](https://ivannaromanovych.github.io/layout_calendar/)


src/index.html
```diff
@@ -7,6 +7,44 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div
+      class="calendar"
+      start-day="sat"
+      month-length="29"
+    >
```
alexander-ignatow: Per task:

> Implement start-day modifier for the calendar with mon, tue, wed, thu, fri, sat and sun values
> Use [@each](https://sass-lang.com/documentation/at-rules/control/each) to create all the modifiers
> The month should start at the correct column (Monday is the 1st, Friday is the 5th)
> You can just add correct margin-left for the first day
> Add a modifier month-length for the calendar with values 28, 29, 30 and 31 (use @for)
> It sets the last day to show (use [nth-child](https://css-tricks.com/how-nth-child-works/))

That means that you are expected to have something like:
```
<div
      class="calendar calendar--start-days--sat calendar--monght-length--29"
 >
```

While what you are using are HTML attributes

Make sure to repeat corresponding topics if you need to.

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  width: $cell-size * 7 + 7;
+  margin: 0 auto;
+}
+
+.calendar {
+  padding: 20px;
+  $root: &;
```
> [BEM] - Create a separate file per each BEM block styles that have the same name as the block
Make sure to repeat the topics on BEM if you need to

Also it makes sense to have separate files for variables and a dedicated main file.

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+    box-sizing: border-box;
+    width: $cell-size;
+    height: $cell-size;
+    border: 1px solid black;
+    font-size: 30px;
+    font-family: Arial, Helvetica, sans-serif;
+    text-align: center;
+    line-height: $cell-size;
+    background-color: #eee;
+    transition: transform 0.5s;
```

> [STYLES] - Make sure to list all styles that you apply transition to.

alexander-ignatow (changes requested):
Some parts of your work is great, but there are some issues.

Most important that you are misusing HTML attributes and BEM

You need to fix these issues to go on!

src/styles/main.scss
ivannaromanovych: I don't quite understand what do I have to change here? Could you please be more specific?

alexander-ignatow: The checklist requires you to apply a transition to all styles you are changing, but check out this:

>     transition: transform 0.5s;

And you are changing this:

 > &:hover {
 >    transform: translateY(-20px);
 >     background-color: #ffbccc;
 > }

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  padding: 20px;
+  $root: &;
```
dffUqp: do you really need this variable?

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  &__content {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 1px;
+    width: $cell-size * 7 + 7;
+    margin: 0 auto;
+  }
+
+  &__day {
+    box-sizing: border-box;
```
dffUqp: it s better to apply this property to all elements (use * selector)
when all elements have the same `box-sizing` value, their behavior becomes more predictable

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    width: $cell-size;
+    height: $cell-size;
+    border: 1px solid black;
+    font-size: 30px;
+    font-family: Arial, Helvetica, sans-serif;
```
dffUqp: let's define a global font-family in `body`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    text-align: center;
+    line-height: $cell-size;
+    background-color: #eee;
+    transition: all 0.5s;
```
dffUqp: fix previous review comment about [transition](https://developer.mozilla.org/en-US/docs/Web/CSS/transition)

dffUqp (changes requested):
LGTM:fire: but needs a few minor fixes:smiley:

image

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--sat calendar--month-length--29">
+      <div class="calendar__content">
```
roman-mirzoian: You do not need to create additional `div` with the `calendar__content` class, it is enough to style the `calendar` class correctly.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+
+    &:hover {
+      transform: translateY(-20px);
+      background-color: #ffbccc;
+    }
+  }
+
+  @each $month-length in $month-lengths {
+    &--month-length--#{$month-length} {
+      @for $i from $month-length +1 through 31 {
```
roman-mirzoian, line 32: Usually you don't put a loop in the class definition, but on the contrary, you make a loop to define many styles for many classes.
That is, now you have a complex system that does not work. It needs to be divided and to think about what is wrong here.

roman-mirzoian (changes requested):
Now your modifiers don't correctly set the number of days and the first day of the week.
image

src/index.html
ivannaromanovych: I don't know how to display a calendar in the middle of the screen while not creating additional div

roman-mirzoian: @ivannaromanovych You need to set the necessary styles for the calendar class.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,11 @@
+$month-lengths: 28, 29, 30, 31;
+$cell-size: 100px;
+$days: (
+  "mon" : 1,
+  "tue" : 2,
+  "wen" : 3,
```
alexander-ignatow: typo!
Wednesday - not Wensday :)

alexander-ignatow (changes requested):
Very good job on the clean-up!

Please fix one last issue that leads to the bug :)

BudnikOleksii (approved):
Nicely done🚀

PR #2494, opened 2023-07-20
add task solution

Vannnkof:
- [DEMO LINK](https://vannnkof.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
   margin: 0;
 }
+
+.page {
+  @extend %positionCenter;
+  min-height: 100vh;
+}
+
+.calendar {
+  max-width: 720px;
```
alexander-ignatow: `max-width` is not the best choice here

image

alexander-ignatow (changes requested):
Good! 
Just minor fix needed

src/styles/main.scss
```diff
@@ -1,3 +1,75 @@
...
 }
+
+.page {
+  @extend %positionCenter;
+  min-height: 100vh;
+}
+
+.calendar {
+  min-width: 720px;
+  width: 720px;
```
BudnikOleksii: The width shouldn't be hardcoded too.
Also, you don't need a min-width if the width is fixed.

BudnikOleksii (changes requested):
Great, but can be even better!
Let's get rid of hardcoded values.

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+
+$start-day: (
+  mon 0,
+  tue 101px,
+  wed 202px,
+  thu 303px,
+  fri 404px,
+  sat 505px,
+  sun 606px
+);
```
roman-mirzoian: It would be cool to have all the variables in a separate file.

src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page">
+    <ul class="calendar calendar--start-day-mon calendar--month-length-31">
+      <li class="calendar__day calendar__day--1"></li>
```
roman-mirzoian: You don't need to create separate classes for each day( `calendar__day--1`, `calendar__day--2`, etc. ), they can be accessed as children of the parent component.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #2500, opened 2023-07-20
layout__calendar

annak1104:


src/styles/main.scss
alexander-ignatow, line 1: Per checklist:
> [BEM] - Create a separate file per each BEM block styles that have the same name as the block

Also it make sense to have separate files for mixins, variables, and a dedicated main file

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
   margin: 0;
 }
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $space-size;
+  width: 720px;
+  margin: 0 auto;
+  padding: 10px;
```
alexander-ignatow: 1. A better idea is not to hardcode the `width` for the calendar but calculate it based on cell width, gaps, padding, etc
2. For some reason actual width of the calendar is `740px` - can you spot why?
image

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+    box-sizing: border-box;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 500ms;
+  }
+
+  &__day:hover {
+    background-color: $color-hover;
+    cursor: pointer;
```
alexander-ignatow: Why alter the cursor property on hover?

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+    transform: translateY(-20px);
+  }
+}
+
+@for $i from 1 through 31 {
+  .calendar__day:nth-child(#{$i})::before {
+    content: "#{$i}";
+    font-size: 30px;
```

> [SASS] - Don't use SASS loops for styles that stay the same for all elements of the group, e.g. display or position.

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
+@mixin cell-propertis ($cell-width, $cell-height, $bg-color) {
+  width: $cell-width;
+  height: $cell-height;
+  background-color: $bg-color;
+}
+
+$space-size: 1px;
```
alexander-ignatow: this variable is used for the border width of the cell and for the gap between cells
It is not a great idea, as most probably these values should not be coupled.

alexander-ignatow (changes requested):
Good solution 👍 

Lets now fix all issues!

src/styles/main.scss
annak1104: I made it according to the instruction.
**On hovering over a cell:

cursor should become pointer
The hovered cell has to become pink**

src/index.html
```diff
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+
+    <div class="calendar calendar--start-day-mon calendar--month-length-31">
```
BudnikOleksii: Remove unnecessary blank line.

src/styles/main.scss
BudnikOleksii: The width is still hardcoded, let's calculate it dynamically.

src/styles/main.scss
BudnikOleksii: > I made it according to the instruction. **On hovering over a cell:
> 
> cursor should become pointer The hovered cell has to become pink**

It's better to apply to the general selector instead of :hover.

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+  padding: 10px;
+
+  &__day {
+    @include cell-propertis(100px, 100px, #eee);
+    border: $space-size solid black;
+    box-sizing: border-box;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition: all 500ms;
```
BudnikOleksii: It's better to specify to what properties you need to apply the transition.
image

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $space-size;
+  width: 720px;
+  margin: 0 auto;
+  padding: 10px;
+
+  &__day {
+    @include cell-propertis(100px, 100px, #eee);
```
BudnikOleksii: 100px looks like a variable. If it was a variable you could dynamically calculate other properties such as the width of the main container.

src/styles/main.scss
BudnikOleksii: It's better to apply the cursor property to the general selector instead of :hover.

src/styles/main.scss
BudnikOleksii: Still not fixed.

src/styles/main.scss
```diff
@@ -1,3 +1,68 @@
...
+    content: "#{$i}";
+    font-size: 30px;
+  }
+}
+
+@each $week-day, $number in $start-day {
+  .calendar--start-day-#{$week-day} >:first-child {
+    margin-left: 100px * ($number - 1) + ($number - 1) * $space-size;
+  }
+}
```
BudnikOleksii: One more proof that 100px should be a variable. If you need to change cell size you should change it here as well.

BudnikOleksii (changes requested):
Nice start👍
Pay attention to comments and fix the code in accordance.

src/styles/variables.scss
```diff
@@ -0,0 +1,23 @@
+@mixin cell-propertis ($cell-width, $cell-height, $bg-color) {
```
roman-mirzoian: Mixins are not variables, so you should keep them in a separate file. 
There is also a typo in the name now (should be `cell-properties`).

src/styles/main.scss
```diff
@@ -1,3 +1,53 @@
...
+  }
+
+  &__day:hover {
+    background-color: $color-hover;
+    transform: translateY(-20px);
+  }
+}
+
+@for $i from 1 through 31 {
+  .calendar__day:nth-child(#{$i})::before {
```
roman-mirzoian: Don't forget about nesting and use it wherever possible.
https://sass-lang.com/guide/#nesting

src/styles/variables.scss
```diff
@@ -0,0 +1,23 @@
...
+  width: $cell-width;
+  height: $cell-height;
+  background-color: $bg-color;
+}
+
+* {
+  box-sizing: border-box;
+}
```
roman-mirzoian: Setting styling via `*` is irrelevant to the variable, so it's best to keep it in a more generic file.

roman-mirzoian (changes requested):

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
+
 body {
   margin: 0;
 }
+
+.calendar {
```
roman-mirzoian, line 17: The styles for the `calendar` class could also be moved to a separate file.

roman-mirzoian (approved):
Well done 👍

PR #2502, opened 2023-07-20
add solution

Yevheniia-Sid:
- [DEMO LINK](https://Yevheniia-Sid.github.io/layout_calendar)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,14 +1,10 @@
-$square-size: 100px;
-$day-of-week: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
-$width-calendar: 728px; //$square-size * 7 + 1(gap) * 8 + 10(padding) * 2
-
 .calendar {
   width: $width-calendar;
   border: 1px solid black;
   display: flex;
   flex-wrap: wrap;
-  gap: 1px;
-  padding: 10px;
+  gap: $gap-day;
+  padding: $padding-calendar;
 
   &__day {
     background-color: #eee;
@@ -18,7 +14,7 @@ $width-calendar: 728px; //$square-size * 7 + 1(gap) * 8 + 10(padding) * 2
     display: flex;
     justify-content: center;
     align-items: center;
-    transition-property: all;
+    transition-property: background-color, transform;
     transition-duration: 0.3s;
 
     &:hover {
@@ -36,7 +32,7 @@ $width-calendar: 728px; //$square-size * 7 + 1(gap) * 8 + 10(padding) * 2
 
   @each $start-day, $i in $day-of-week {
     &--start-day--#{$start-day} :nth-child(1) {
-      margin-left: #{$i*101}px;
+      margin-left: #{$i*$margin-left};
     }
   }
 
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,50 @@
+$square-size: 100px;
+$day-of-week: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+
+.calendar {
+  width: 710px;
+  border: 1px solid black;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 10px;
+
+  &__day {
+    background-color: #eee;
+    width: $square-size;
+    height: $square-size;
+    box-sizing: border-box;
+    border: 1px solid #000;
+    font-size: 30px;
+    font-family: Arial, Helvetica, sans-serif;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-duration: 0.5s;
+
+    &:hover {
+      cursor: pointer;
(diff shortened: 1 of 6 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
+$square-size: 100px;
+$day-of-week: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+
+.calendar {
+  width: 710px;
```
Anna-Tkachenko: Will be better to move the with to separate variable, for not it's some magic number

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    width: $square-size;
+    height: $square-size;
+    box-sizing: border-box;
+    border: 1px solid #000;
+    font-size: 30px;
+    font-family: Arial, Helvetica, sans-serif;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-duration: 0.5s;
```
Anna-Tkachenko, line 18: Let's specify the property for which you want to set up transition duration value

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 10px;
+
+  &__day {
+    background-color: #eee;
+    width: $square-size;
+    height: $square-size;
+    box-sizing: border-box;
```
Anna-Tkachenko: It's the general property that will be applied for all elements inside the project, so let's set it using `*` selector

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    border: 1px solid #000;
+    font-size: 30px;
+    font-family: Arial, Helvetica, sans-serif;
```
Anna-Tkachenko: Move basic font styles to `html` tag

Anna-Tkachenko (changes requested):

src/styles/calendar.scss
```diff
@@ -0,0 +1,49 @@
+$square-size: 100px;
+$day-of-week: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+$width-calendar: 728px; //$square-size * 7 + 1(gap) * 8 + 10(padding) * 2
```
Moroz-Dmytro: Create a file for variables and let's not hardcode width but do calculation that are commented

src/styles/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  &__day {
+    background-color: #eee;
+    width: $square-size;
+    height: $square-size;
+    border: 1px solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-property: all;
```
Moroz-Dmytro: Let's not use `all`, write here styles to which you apply transition. For example `color`

src/styles/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  @for $number from 1 through 31 {
+    & :nth-child(#{$number})::before {
+      content: "#{$number}";
+    }
+  }
+
+  @each $start-day, $i in $day-of-week {
+    &--start-day--#{$start-day} :nth-child(1) {
+      margin-left: #{$i*101}px;
```
Moroz-Dmytro: Why we * to 101? Create variable for this value that will describe it

src/styles/main.scss
```diff
@@ -1,3 +1,17 @@
...
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
+
 body {
   margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
```
Moroz-Dmytro: Better will be to move this to reset/normalize files

Moroz-Dmytro (changes requested):
Good progress 👍 
Some advice below 😉

src/styles/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+  &__day {
+    background-color: #eee;
+    width: $square-size;
+    height: $square-size;
+    border: 1px solid #000;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transition-property: background-color, transform;
+    transition-duration: 0.3s;
```
alexander-ignatow: Per task:
> All changes should be animated with the duration of 0.5s

src/styles/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+
+    &:hover {
+      cursor: pointer;
```
alexander-ignatow: No need to change `cursor` on hover - a better idea set it as a pointer to the elements regardless of the hover state

alexander-ignatow (changes requested):
Almost perfect - just cover two minor issues!

BudnikOleksii (approved):
Looks great🔥

PR #2505, opened 2023-07-21
Develop

neovendano:
- [DEMO LINK](https://neovendano.github.io/layout_calendar/)
- [TEST REPORT LINK](https://neovendano.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-thu calendar--month-length-29">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
...
+    align-items: center;
+    justify-content: center;
+    background-color: $background-gray;
+
+    width: $day-width;
+    height: $day-width;
+
+    border-width: 1px;
+    border-style: solid;
+    border-color: black;
```
alexander-ignatow: Hint: it is often simpler use `border` shorthand here
```
border: 1px solid black;
```

src/styles/main.scss
```diff
@@ -1,3 +1,61 @@
...
+
+    transition: transform 0.5s, background-color 0.5s;
+
+    &:hover {
+      cursor: pointer;
```
alexander-ignatow, line 34: Why change the cursor style on hover?

src/styles/variables.scss
```diff
@@ -0,0 +1,12 @@
...
+  mon 0,
+  tue 1,
+  wed 2,
+  thu 3,
+  fri 4,
+  sat 5,
+  sun 6;
+$day-width: 100px;
+$background-gray: #eee;
+$background-pink: pink;
```
alexander-ignatow: Avoid using color names in color variables

Consider using background, cell-background, background-hover, etc

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-thu calendar--month-length-29">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
```
alexander-ignatow, line 14: consider using semantic tags for calendar and cells

alexander-ignatow (changes requested):
Strong 👍 

Just fix a few minor issues, please!

src/styles/main.scss
neovendano: It`s a condition of the task:

On hovering over a cell:
- cursor should become pointer
- The hovered cell has to become pink

alexander-ignatow: Good point, but if you set `cursor: pointer;` to any cell, you will see the cursor only on the hovered one :)

On the other hand - with your solution you unnecessarily altering `cursor` property on hover

Try changing it and see if there is any difference with your solution.

alexander-ignatow (changes requested):
Very good!

Let's fix the last remaining issue

Anna-Tkachenko (approved):
Good job!

PR #2507, opened 2023-07-21
calendar created

rostyslav48:


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+    align-items: center;
+    background-color: #eee;
+    width: 100px;
+    height: 100px;
+    border: solid 1px #000;
+    box-sizing: border-box;
+    transition: transform 0.5s;
+
+    &:hover {
+      cursor: pointer;
```
alexander-ignatow: Why are you altering `cursor` here?

alexander-ignatow (changes requested):
Very good 👍 

A minor issue is needed to be fixed.

src/styles/blocks/calendar.scss
rostyslav48: When you are pointing to the day it becomes pink and moves 20px up + cursor becomes pointer in order to make user experience even better (without this feature browser treat each day as a text and offers to copy it)

rostyslav48: Oh no actually it works well even with a common cursor, however I like more when it is pointer, do I have to change it?

rostyslav48: Anyway, I have already changed it

src/index.html
```diff
@@ -7,6 +7,40 @@
...
+        <li class="calendar__day calendar__day--22"></li>
+        <li class="calendar__day calendar__day--23"></li>
+        <li class="calendar__day calendar__day--24"></li>
+        <li class="calendar__day calendar__day--25"></li>
+        <li class="calendar__day calendar__day--26"></li>
+        <li class="calendar__day calendar__day--27"></li>
+        <li class="calendar__day calendar__day--28"></li>
+        <li class="calendar__day calendar__day--29"></li>
+        <li class="calendar__day calendar__day--30"></li>
+        <li class="calendar__day calendar__day--31"></li>
```
BudnikOleksii: You can use :nth-child and get rid of all unnecessary calendar__day--31 modifiers😉

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,63 @@
+$cell-side: 100px;
+$cell-sideAndGap: $cell-side + 1;
+$weekDays: (
+  "mon": 0,
+  "tue": $cell-sideAndGap * 1,
+  "wed": $cell-sideAndGap * 2,
+  "thu": $cell-sideAndGap * 3,
+  "fri": $cell-sideAndGap * 4,
+  "sat": $cell-sideAndGap * 5,
+  "sun": $cell-sideAndGap * 6
```
BudnikOleksii: You can do these calculations in the @each loop where you set margins.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #eee;
+    width: 100px;
+    height: 100px;
```
BudnikOleksii: You have a $cell-side variable

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,63 @@
...
+    border: solid 1px #000;
+    box-sizing: border-box;
+    transition: transform 0.5s;
```
BudnikOleksii: As I see in hover you change not only a transform property, let's add a transition to background-color too

BudnikOleksii (changes requested):
Good, but can be even better!
Check the comments.

src/styles/blocks/calendar.scss
rostyslav48: Thanks for your review, have already done all changes)

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <section class="calendar calendar__start-day--mon calendar__month-length--31">
```
roman-mirzoian: You don't need to wrap everything in an additional section.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
+$cell-side: 100px;
+$cell-sideAndGap: $cell-side + 1;
+$weekDays: ("mon", "tue", "wed", "thu", "fri", "sat", "sun");
```
roman-mirzoian: It is better to put all the variables in a separate file.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
...
+    box-sizing: border-box;
+    transition: transform, background-color, 0.5s;
+
+    &:hover {
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+
+  $counter: 0;
```
roman-mirzoian: It is better to keep all the variables together and not spread them over the file.

It is also necessary to give more meaningful names to variables, for example `daysCounter`.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
+$cell-side: 100px;
+$cell-sideAndGap: $cell-side + 1;
```
roman-mirzoian: Variables should be written in the same style: `cell-side-and-gap` or `cellSideAndGap`

roman-mirzoian (changes requested):

src/styles/blocks/calendar.scss
rostyslav48: This direct variable refers only to the loop under it, so it might cause some misunderstanding if I move it upper, shall I do it?

roman-mirzoian: @rostyslav48 It is worth giving a correct name for this variable and then there will be no misunderstandings

rostyslav48: Ok, but may I leave it on the same place?

rostyslav48:
I have applied all the changes exept moving the var daysCount somewhere. This var is used with the cycle below only, and I don't see any rational point of moving it to the top of the document.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,49 @@
+.calendar {
+  display: flex;
+  gap: 1px;
+  padding: 10px;
+  box-sizing: border-box;
```
dffUqp, line 5: it s better to apply this property to all elements (use * selector)

when all elements have the same `box-sizing` value, their behavior becomes more predictable

dffUqp (approved):
LGTM🔥

PR #2516, opened 2023-07-21
add task solution

valet2102:
- [DEMO LINK](https://valet2102.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,55 @@
+@import "./reset";
+@import "./_variables";
+
 body {
-  margin: 0;
+  font-family: Arial, sans-serif;
+  font-size: 30px;
+}
+
+.wrapper {
+  display: flex;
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+    align-items: center;
+    background-color: $cell-bg-color;
+    transition: all 0.3s;
+
+    @media (any-hover: hover) {
+      &:hover {
+        transform: translateY(-20px);
+        background-color: $cell-bg-color-hover;
+      }
+    }
```
alexander-ignatow, line 37: 👍  for digging into a hover media query

However, in the current task, you don't need to use it, as the hover effect does not affect the usability on non-hover devices

No need to fix this - just keep in mind for the next time

alexander-ignatow (approved):
Very good 👍 

LGTM but see the comment

PR #2520, opened 2023-07-21
add task solution

Wesses:
- [DEMO LINK](https://Wesses.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Wesses.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,38 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <ul class = "calendar calendar--first-day--sun calendar--month-length--31">
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
-      <li class = "calendar__day"></li>
+    <ul class="calendar calendar--first-day--sun calendar--month-length--31">
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
+      <li class="calendar__day"></li>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <ul class = "calendar calendar--first-day--sun calendar--month-length--31">
```
Anna-Tkachenko: Nice to see semantic tags usage ⭐️

src/index.html
```diff
@@ -7,6 +7,38 @@
...
+      <li class = "calendar__day"></li>
+      <li class = "calendar__day"></li>
```
Anna-Tkachenko: The space before and after `=` sign is redundant, fix it
```suggestion
      <li class="calendar__day"></li>
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+.calendar {
+  margin: 0;
```
Anna-Tkachenko: We can't set margins to bem block 🐝

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  padding: 30px 10px;
+  width: $calendar-width;
+
+  list-style-type: none;
+
+  &__day {
+    box-sizing: border-box;
```
Anna-Tkachenko, line 11: It's the general rule that will be applied to the all elements inside the project, will be better to set it using `*` selector

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+    width: $square-side;
+    height: $square-side;
+    border: 1px solid black;
+    text-align: center;
+    line-height: $square-side;
+    background-color: #eee;
+
+    transition: transform 0.5s;
+
+    cursor: default;
```
Anna-Tkachenko: Do you need this default property?

src/styles/blocks/default-styles.scss
```diff
@@ -0,0 +1,6 @@
+body {
+  margin: 0;
+
+  font-size: 30px;
```
Anna-Tkachenko, line 2: Let's move basic font styles to `html` tag

src/styles/blocks/variables.scss
```diff
@@ -0,0 +1,11 @@
+$square-side: 100px;
```
Anna-Tkachenko: It's not clear what is `square` in this task
Probably will be better to specify the name relatively to the `date` naming

Anna-Tkachenko (changes requested):

src/styles/blocks/calendar.scss
Wesses: Its not, if you talk about box-sizing. It only applies to day-boxes(class="calendar__day, but not to the container(class="calendar").

Anna-Tkachenko: Agree, as it's how it works for now. But generally, when you'll use it in the large projects, you'll specify it as the general rule for all elements, as you should decide how you'll calculate the size for the all elements
So this property should work the same for all elements on your page

src/styles/blocks/variables.scss
```diff
@@ -0,0 +1,11 @@
+$date-box-side: 100px;
+$calendar-width: (7 * $date-box-side + 6px);
+$week-days: (
+  "mon" ($date-box-side * 0),
+  "tue" ($date-box-side + 1px),
+  "wed" ($date-box-side * 2 + 2px),
+  "thu" ($date-box-side * 3 + 3px),
+  "fri" ($date-box-side * 4 + 4px),
+  "sat" ($date-box-side * 5 + 5px),
+  "sun" ($date-box-side * 6 + 6px)
```
Anna-Tkachenko, line 10: Will be better to make these calculations in the loop, as you already have these values (0-6) there
In case if you need to change the logic, it will be easier to change it in one place in the loop, instead of refactoring the values for all this variables

Anna-Tkachenko (approved):

src/styles/blocks/variables.scss
Wesses: Thanks for your edits, I will definitely take them into account in future projects.

PR #2526, opened 2023-07-22
Calendar

vatatan:
- [DEMO LINK](https://vatatan.github.io/layout_calendar/)
- [TEST REPORT LINK](https://vatatan.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -3,12 +3,10 @@
   flex-wrap: wrap;
   gap: $cell-space;
   width: $cal-width;
-  padding: 20px;
-  margin: 0 auto;
+  padding: $cal-padding;
 
   &__day {
     display: flex;
-    box-sizing: border-box;
     align-items: center;
     justify-content: center;
     cursor: pointer;
@@ -16,10 +14,10 @@
     height: $cell-size;
     width: $cell-size;
     background-color: $bg-color-day;
+    transition-duration: 0.5s;
 
     &:hover {
       background-color: $hover-cell-color;
-      transition-duration: 0.5s;
       transform: translate(0, -20px);
     }
 
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,46 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $cell-space;
+  width: $cal-width;
+  padding: 20px;
+  margin: 0 auto;
+
+  &__day {
+    display: flex;
+    box-sizing: border-box;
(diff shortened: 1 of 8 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $cell-space;
+  width: $cal-width;
+  padding: 20px;
+  margin: 0 auto;
```
Anna-Tkachenko: You can't set margins to BEM block 🐝

src/styles/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+
+  &__day {
+    display: flex;
+    box-sizing: border-box;
```
Anna-Tkachenko: it's the general rule, that will be used for all elements inside the project, so will be better to set it using `*` selector

src/styles/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+    justify-content: center;
+    cursor: pointer;
+    border: 1px solid black;
+    height: $cell-size;
+    width: $cell-size;
+    background-color: $bg-color-day;
+
+    &:hover {
+      background-color: $hover-cell-color;
+      transition-duration: 0.5s;
```
Anna-Tkachenko: Check point 9 in checklist

Anna-Tkachenko (changes requested):

src/styles/calendar.scss
vatatan: done

src/styles/calendar.scss
vatatan: done

src/styles/calendar.scss
vatatan: done

src/styles/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+  &__day {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    cursor: pointer;
+    border: 1px solid black;
+    height: $cell-size;
+    width: $cell-size;
+    background-color: $bg-color-day;
+    transition-duration: 0.5s;
```
Anna-Tkachenko: Let's specify the property, for which you'll use transition duration
If you don't specify the property, it will affect in a bad way on the website performance

Anna-Tkachenko (approved):
⭐️

src/styles/calendar.scss
vatatan: done

PR #2529, opened 2023-07-23
calendar task

olegobiukh:
[DEMO](https://olegobiukh.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -25,6 +25,7 @@ body {
   }
   &__item {
     @include content-center;
+    display: none;
     width: $size;
     height: $size;
     background-color: #eee;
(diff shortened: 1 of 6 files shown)


SanyaBratashchuk (changes requested):
image
Implement it

etojeDenys (changes requested):
I should be able to change the number of cells by changing the class modifier

olegobiukh:
> I should be able to change the number of cells by changing the class modifier

Thank you so much for your description. Fixed.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,64 @@
...
+        content: "#{$day}";
+      }
+    }
+  }
+  &__item {
+    @include content-center;
+    display: none;
+    width: $size;
+    height: $size;
+    background-color: #eee;
```
loralevitska, line 31: It is a good practice to make all colors variable

loralevitska (approved):
Good job!

PR #2530, opened 2023-07-23
Develop

00Mass00:
- [DEMO LINK](https://00Mass00.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,7 @@
+.container {
+  display: flex;
+  justify-content: center;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+  padding-top: 200px;
+}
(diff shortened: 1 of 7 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,7 @@
+.container {
+  display: flex;
+  justify-content: center;
+  font-family: Arial, Helvetica, sans-serif;
```
Anna-Tkachenko, line 4: Will be better to move basic font styles to `html` tag

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $full-width;
+  gap: $space;
+  padding: $padding;
+
+  &__day {
+    display: flex;
+    box-sizing: border-box;
```
Anna-Tkachenko, line 10: Will be better to specify this property via `*` selector as it will be applied as the general rule for all elements in your project

Anna-Tkachenko (approved):
Good job!

PR #2531, opened 2023-07-23
create calendar

mishakoshilka:
- [DEMO LINK](https://mishakoshilka.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


anastasiiavorobiova (changes requested):
Great work! 1) The modifier for the first day should be on the calendar class. It should work dynamically while changing it in the browser: consider fixing it. 2) Consider centering your calendar on the page

anastasiiavorobiova (approved):
Awesome!

PR #2532, opened 2023-07-23
finish soulution

serkrops:


src/styles/main.scss
```diff
@@ -1,3 +1,60 @@
-body {
+$cell-size: 100px;
+$week: mon 1, tue 2, wen 3, thu 4, fri 5, sat 6, sun 7;
```
roman-mirzoian: Better to keep variables in a separate file.

roman-mirzoian (changes requested):
Now the calendar does not start from the day specified in the modifier:
image
And if you specify Sunday, it will generally break:
image

roman-mirzoian (changes requested):
Still not working.
image

serkrops:
> 

try again pls

roman-mirzoian (approved):
Well done 👍

PR #2533, opened 2023-07-23
add task solution layout_calendar

thaiduk:
- [DEMO LINK](https://thaiduk.github.io/layout_calendar/)
- [TEST REPORT LINK](https://thaiduk.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -25,21 +25,20 @@
       }
     }
 
-    @for $last-day from 28 through 31 {
-      &--month-length--#{$last-day} > :not(:nth-child(-n + #{$last-day})) {
-        display: none;
-      }
-    }
-
-    @each $start-day, $number in $start-days {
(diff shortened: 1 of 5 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,46 @@
+.calendar {
+  box-sizing: border-box;
+  width: calc(#{$boxSize} * 7 + #{$gapSize} * 6 + 20px);
```
roman-mirzoian: Mathematical calculations can also be transferred to a variable and use the finished result.

roman-mirzoian (changes requested):
Now it turns out that the week shows six days instead of seven:
image
Also, when changing the total number of days and setting the first day of the week, nothing changes:
image

alexander-ignatow (approved):
Good! : 👍

PR #2534, opened 2023-07-23
add task solution

AngerDESTROYS:
- [DEMO LINK](https://AngerDESTROYS.github.io/layout_calendar/)
- [TEST REPORT LINK](https://AngerDESTROYS.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
 body {
   margin: 0;
+  @import "variables";
```
roman-mirzoian: All imports should be made at the beginning of the file and they should not be placed in specific tags.

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+      cursor: pointer;
+
+      &:hover {
+        transform: translate(0, -20px);
+        background-color: $calendar-color-hover;
+      }
+    }
+
+    @for $i from 0 through 31 {
+
```
roman-mirzoian: Do not leave extra empty lines.

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+      &__day:nth-child(#{$i + 1}):before {
+        content: "#{$i + 1}";
+      }
+    }
+
+    @for $length from 28 through 31 {
+      &--month-length--#{$length} :not(:nth-child(-n + #{$length})) {
+        opacity: 0;
+        pointer-events: none;
```
roman-mirzoian: There is no need to set styles in a loop that can be set once for a class.

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+      }
+    }
+  }
+
+  @each $day, $num in $days-of-week {
+    .calendar--day-start--#{$day} :first-child {
```
roman-mirzoian: Don't forget to use nesting.

src/styles/variables.scss
```diff
@@ -0,0 +1,6 @@
+$gap-between-days: 1px;
+$days-height-width: 100px;
```
roman-mirzoian, line 2: It is better not to create shared variables for different parameters.

roman-mirzoian (changes requested):
Also, now you have an unnecessary scroll that needs to be removed:
https://gyazo.com/2e867506ecdaf17920d144dcb5fa0c9c

src/styles/calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+  @for $i from 0 through 31 {
+    &__day:nth-child(#{$i + 1}):before {
+      content: "#{$i + 1}";
+    }
+  }
+
+  @for $length from 28 through 31 {
+    &--month-length--#{$length} :not(:nth-child(-n + #{$length})) {
+      opacity: 0;
+      pointer-events: none;
```
alexander-ignatow, line 38: Simple `display: none` is probably a better option here

src/styles/calendar.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+  position: relative;
+  display: flex;
+  align-items: center;
+  margin: 150px auto;
```
alexander-ignatow, line 7: No margins on BEM blocks

src/styles/normalize.scss
```diff
@@ -0,0 +1,7 @@
+* {
+  box-sizing: border-box;
+}
+
+.body {
+  margin: 20px 0 0;
```
alexander-ignatow, line 6: In most cases it is impractical setting margins on `body` 
It is better to reset them and use margins/paddings with inner containers

alexander-ignatow (approved):
Your solution works - but please review all comments

Such comments may lead to the rejection with next tasks

PR #2538, opened 2023-07-24
add task solution

SmartSergiZ:
[DEMO LINK](https://SmartSergiZ.github.io/layout_calendar/)
[TEST REPORT LINK](https://SmartSergiZ.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -12,7 +12,6 @@ body {
 .calendar {
   display: flex;
   flex-wrap: wrap;
-  justify-content: center;
   gap: $gap-size;
   width: $ceil-width * 7 + $gap-size * 6 + 20px;
   padding: 20px 10px 10px;
@@ -29,7 +28,7 @@ body {
     transition: transform 0.5s;
     cursor: pointer;
 
-    @for $i from 1 through 29 {
+    @for $i from 1 through 31 {
       &:nth-child(#{$i})::before {
         content: "#{$i}";
       }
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+$month-length: 28, 29, 30, 31;
+
+html {
+  font-size: 30px;
   margin: 0;
+  padding: 0;
+}
+
+.calendar {
+  margin: 0 auto;
```
BudnikOleksii: BEM-block shouldn't have any margins, it you want to specify them use BEM mix instead.

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+    box-sizing: border-box;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #eee;
+    border: 1px solid black;
+    width: $ceil-width;
+    height: $ceit-height;
+    font-size: 30px;
+    transition: all 0.5s;
```
BudnikOleksii: It's better to specify what properties should apply the transition.
image

BudnikOleksii (changes requested):
Great work!
Left a few comments, check them.
Also, let's move variables to separate file.
And in accordance with task add cursor pointer:
image

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+  &__cell {
+    box-sizing: border-box;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #eee;
+    border: 1px solid black;
+    width: $ceil-width;
+    height: $ceit-height;
+    font-size: 30px;
```
alexander-ignatow: You already have `font-size` in the body - why duplicate it here?

In addition: per task:
> Add a number (Arial 30px)

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+    transition: transform 0.5s;
+
+    @for $i from 1 through 29 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+      }
+    }
+
+    &:hover {
+      cursor: pointer;
```
alexander-ignatow: The `cursor:pointer` is ok  👍 , but why are you setting it on hover?

alexander-ignatow (changes requested):
Good job fixing comments - still some are left.

Let's fix them!

roman-mirzoian (changes requested):
Almost done, but now the number of days doesn't change after updating the days modifier and the cursor doesn't change when hovering over a cell.
https://gyazo.com/0d8db61bf63b733b629be87e526b5167
image

roman-mirzoian (approved):
Well done 🔥

PR #2539, opened 2023-07-24
add task solution

oleg-matvienko:
https://oleg-matvienko.github.io/layout_calendar/

### DIFF
--- a/src/styles/_calendar.scss
+++ b/src/styles/_calendar.scss
@@ -8,6 +8,7 @@
   gap: $gap-size;
 
   &__day {
+    font-family: $font;
     display: flex;
     justify-content: center;
     align-items: center;
--- a/src/styles/_calendar.scss
+++ b/src/styles/_calendar.scss
@@ -1,5 +1,4 @@
 .calendar {
-  margin: 0 auto;
   box-sizing: border-box;
   display: flex;
   width: $cell-size * 7 + $gap-size * 6 + 20px;
@@ -8,18 +7,18 @@
   gap: $gap-size;
 
   &__day {
-    font-family: $font;
     display: flex;
-    justify-content: center;
-    align-items: center;
     box-sizing: border-box;
     width: $cell-size;
     height: $cell-size;
+    justify-content: center;
+    align-items: center;
     border: 1px solid black;
     list-style: none;
     background-color: #eee;
-    transition: color 500ms, transform 500ms;
+    transition: background-color 500ms, transform 500ms;
     cursor: pointer;
+    font-size: 30px;
 
     @for $n from 1 through 31 {
       &:nth-child(#{$n})::before {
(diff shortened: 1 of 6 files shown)


src/styles/_page.scss
```diff
@@ -0,0 +1,4 @@
+.page {
```
roman-mirzoian: I don't see that you use such a class in html, where did you get it from?

src/styles/main.scss
```diff
@@ -1,3 +1,21 @@
+@import "variables.scss";
+@import "calendar.scss";
+@import "page.scss";
+
```
roman-mirzoian: You do not use tags `h1` - `h6`, `p`, `ol` in your HTML. 
There is no need to write styles for things that are not used in the code.

roman-mirzoian (changes requested):
The calendar works fine, but I see you have styles for elements that are not used in your code, why did you decide to write those styles?

src/styles/_page.scss
oleg-matvienko: I get it now. It's misunderstanding, I used my project in past like notebook and forget remove.

alexander-ignatow (changes requested):
Good!

But you are using wrong font, per task should be:
> Arial 30px

src/styles/_calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    width: $cell-size;
+    height: $cell-size;
+    border: 1px solid black;
+    list-style: none;
+    background-color: #eee;
+    transition: color 500ms, transform 500ms;
+    cursor: pointer;
```
Moroz-Dmytro: Remember about [css order](https://github.com/necolas/idiomatic-css#declaration-order)

src/styles/_calendar.scss
```diff
@@ -0,0 +1,47 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    width: $cell-size;
+    height: $cell-size;
+    border: 1px solid black;
+    list-style: none;
+    background-color: #eee;
+    transition: color 500ms, transform 500ms;
```
Moroz-Dmytro: transition should be for background-color, not for color

src/styles/_calendar.scss
```diff
@@ -0,0 +1,47 @@
+.calendar {
+  margin: 0 auto;
```
Moroz-Dmytro: Don't add margins to BEM block

Moroz-Dmytro (changes requested):
Still something wrong with font
image

alexander-ignatow (approved):
Let's move on - but the next time make sure to deploy the latest version before submitting the task (your deploy still has the wrong font)

PR #2542, opened 2023-07-24
task solution

denlubn:
- [DEMO LINK](https://denlubn.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,57 @@
+@import "./variables.scss";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,57 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
 }
+
+.calendar {
```
vlad-demchuk, line 11: create separate files for bem blocks and reset

vlad-demchuk (changes requested):
GJ, just a small improvement

PR #2545, opened 2023-07-25
add task solution

Aserg1992:
[DEMO LINK](https://Aserg1992.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,48 @@
+.calendar {
+  margin: 0 auto;
+  padding: 10px;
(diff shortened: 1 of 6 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,48 @@
+.calendar {
+  margin: 0 auto;
```
BudnikOleksii, line 2: BEM-block shouldn't have any margins, it you want to specify them use BEM mix instead.

BudnikOleksii (approved):
Nicely done🔥
Just don't use margins for the BEM block, use mix instead if you need. BEM blocks should be independent and you can reuse them in every part of your page.

PR #2546, opened 2023-07-26
calendar solution

cherniavskaaa:


src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
 body {
+  font-family: $number-font;
+  font-size: $number-size;
   margin: 0;
 }
+
+.calendar {
+  display: flex;
+  box-sizing: border-box;
+  margin: 0 auto;
```
Moroz-Dmytro, line 10: Don't set external styles to the block. [Typical BEM mistakes](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes)

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+    @for $day from 1 through 31 {
+      &:nth-child(#{$day}) {
+        &::before {
+          content: "#{$day}";
+        }
+      }
+    }
+
+    &:hover {
+      cursor: pointer;
```
Moroz-Dmytro, line 38: We don't need to set this again

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$number-font: "arial";
+$number-size: 30px;
+
+$cell-size: 100px;
+$gap-size: 1px;
+$calendar__height: 20px;
+
+$calendar-size: $cell-size * 7 + $gap-size * 6;
+$month-lengths: 28, 29, 30, 31;
+$start-days: "mon" 0, "tue" 1, "wed" 2, "thu" 3, "fri" 4, "sat" 5, "sun" 6;
```
Moroz-Dmytro, line 15: Some of these values are not using. Use them or delete

Moroz-Dmytro (changes requested):
Some work still needs to be done. Check the comments below. Also please add a link to the demo
Looks like modifiers don't work(I want to see 28 days, starting since wed)

image

Also
image

PR #2552, opened 2023-07-28
add solution

Daniil-A4:
- [DEMO LINK](https://Daniil-A4.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,45 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gapValue;
+  padding: 10px;
+  width: calc((7 * #{$side}) + (#{$gapValue} * 6));
+
+  @for $i from 1 through 31 {
+    &__day:nth-child(#{$i})::before {
+      content: "#{$i}";
+    }
+  }
+
+  &__day {
+    background-color: #eee;
+    width: $side;
+    height: $side;
+    text-align: center;
+    line-height: $side;
+    border: 1px solid black;
+    box-sizing: border-box;
+    font-size: 30px;
+    cursor: pointer;
+    transition: transform 0.5s, background-color 0.5s;
+
+    &:hover {
+      background-color: pink;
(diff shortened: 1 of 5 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+    &--start-day--#{$day} :first-child {
+      margin-left: $value * ($side + $gapValue);
+    }
+  }
+
+  @for $month-length from 28 through 31 {
+    &--month-length--#{$month-length} :nth-child(n + #{$month-length + 1}) {
+      display: none;
+    }
+  }
```
andriy-shymkiv, line 42: something wrong with this loop because I see 29,30,31 day in your demo

src/styles/calendar.scss
```diff
@@ -0,0 +1,45 @@
...
+    cursor: pointer;
+    transition: transform 0.5s, background-color 0.5s;
+
+    &:hover {
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+
+  $days: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
```
andriy-shymkiv: move to variables dir

src/styles/main.scss
```diff
@@ -1,3 +1,14 @@
...
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
 }
+
+.wrap {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
```
andriy-shymkiv: do you really need this wrapper? I mean you can give the same styles to `calendar` block

andriy-shymkiv (changes requested):
good job! left comments

src/styles/main.scss
Daniil-A4: @Andriy-Shymkiv i can't move this styles to calendar block because of markup will be broken, i added margin: 0 auto but it isn't good idea, so which option i can use here?

denys-danyliuk (approved):

PR #2562, opened 2023-07-31
add task solution

UlianaChorna:
- [DEMO LINK](https://UlianaChorna.github.io/layout_calendar/)
- [TEST REPORT LINK](https://UlianaChorna.github.io/layout_calendar/report/html_report/)


Esceype (approved):
One step to perfect work, that`s about file structure, better to move all blocks styling to some folder and variables to another one, same about the body, which should be normalize.

PR #2573, opened 2023-08-03
done

ikispower:
[DEMO LINK] https://ikispower.github.io/layout_calendar/


volodymyr-soltys97 (changes requested):
When I choose `mon` in the modifier I get this problem. There should always be 7 columns
image

nazarmatsevych (approved):
Good job!

PR #2575, opened 2023-08-06
done

no4kar:
- [DEMO LINK](https://no4kar.github.io/layout_calendar/)

Hi.
When I try to split a calendar__item--modifiers into separate rows as here
```
<div class="
  calendar__item
  calendar__item--start-day-wed
  calendar__item--month-length-29"
>
```
,linter throws "11:7  error  Tag can't have more than one block [] (one-block)" and "13:9  error  Element isn't inside the parent block  [calendar__day]  (element-inside-parent-block)".
How solve it

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <div class="calendar calendar--start-day-sat calendar--month-length-31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+      </div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <div class="calendar calendar--start-day-sat calendar--month-length-31">
```
mykhalenych: https://mate-academy.github.io/fe-program/css/typical-bem-mistakes

pay attention to modifiers

mykhalenych (changes requested):

src/index.html
```diff
@@ -7,6 +7,40 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
+      <div class="calendar__item calendar__item--start-day-wed calendar__item--month-length-29">
```
Viktor-Kost: it'd be better to follow BEM recommendations like this
```suggestion
    <div class="container">
      <div class="calendar calendar--start-day--wed calendar--month-length--29">
```

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  + ($calendar-gap * $calendar-columns)
+  + ($calendar-padding * 2);
+
+* {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: Arial, sans-serif;
+  font-size: 30px;
+}
```
Viktor-Kost: it's not a good practice to use styling with * selector. It's allowed to use only box-sizing: border-box at this selector

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+      }
+    }
+  }
+
+  &__day {
+    position: relative;
+    width: $day-cell-size;
+    border: $day-border-width solid black;
+    background-color: $day-bckgrd-color;
+    transition: all 0.5s;
```
Viktor-Kost: specify what rules you'd like to change
```suggestion
    transition: background-color 0.5s, transform 0.5s;
```

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+
+    &::before {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      line-height: calc(#{$day-cell-size} - (#{$day-border-width} * 2));
```
Viktor-Kost: no needs to calculate a value for line-height. Just use a simple value

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+
+  &__item {
+    display: flex;
+    flex-wrap: wrap;
+    gap: $calendar-gap;
+    width: $calendar-width;
+    min-width: $calendar-width;
```
Viktor-Kost: redundant since you've set width
```suggestion
```

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

src/styles/main.scss
no4kar: in previous advises was "avoid hardcoding"

src/index.html
no4kar: but it was from the beginning. the previous mentor rejected and demanded a replacement.

denys-danyliuk (approved):

PR #2592, opened 2023-08-10
Develop

patrykyo20:
https://Patrykmclaren14.github.io/layout_calendar/
https://Patrykmclaren14.github.io/layout_calendar/report/html_report/


src/index.html
```diff
@@ -4,9 +4,49 @@
...
+        <div class="block block--22"></div>
+      </div>
+      <div class="calendar__block">
+        <div class="block block--23"></div>
+        <div class="block block--24"></div>
+        <div class="block block--25"></div>
+        <div class="block block--26"></div>
+        <div class="block block--27"></div>
+        <div class="block block--28"></div>
+        <div class="block block--29"></div>
```
VitaliyBodnar: you don't need modificator  for every blocks, it's supposed be only for calendar blocks parrent

VitaliyBodnar (changes requested):

src/index.html
```diff
@@ -4,9 +4,43 @@
...
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+        <div class="calendar__row-block"></div>
+      </div>
+    </div>
```
JoyCoffeeAddict: You should have one div for calendar, and 31 divs inside for each day 
`calendar__row` is unnecessary. Also both modifiers should be on `calendar` and not on individual day.
Also change your classes to match readme and checklist.

src/styles/blocks/body.scss
JoyCoffeeAddict, line 1: You don't need additional file just for `body`

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,41 @@
...
+@use "../utils/placeholders";
+
+.calendar {
+  @extend %center;
+  height: 100vh;
+
+  &__row {
+    display: flex;
+    flex-wrap: wrap;
+    width: 750px;
```
JoyCoffeeAddict: You have forgotten about the gap, also consider it when calculating the margins for moving the first day of the month

JoyCoffeeAddict (changes requested):

mbruhler (approved):
Well done! Keep up the good work.

In the future remember to use proper English words (length not lenght).

PR #2603, opened 2023-08-14
calendar

RomanO3GIT:
[DEMO LINK](https://RomanO3GIT.github.io/layout_calendar/)

### DIFF
--- a/src/main.css
+++ b/src/main.css
@@ -1,0 +1,196 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
+
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: 710px;
+  gap: 1px;
+}
+
+.calendar__date {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 100px;
+  height: 100px;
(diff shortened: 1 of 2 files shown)


src/main.css
```diff
@@ -0,0 +1,196 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
```
loralevitska, line 5: Only `box-sizing` property allows for `*` selector. Move other properties to html or body tag

src/main.css
```diff
@@ -0,0 +1,196 @@
...
+  gap: 1px;
+}
+
+.calendar__date {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 100px;
+  height: 100px;
+  background-color: #eee;
```
loralevitska, line 28: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/main.css
```diff
@@ -0,0 +1,196 @@
...
+
+.calendar__date:nth-child(2):before {
+  content: "2";
+}
+
+.calendar__date:nth-child(3):before {
+  content: "3";
+}
+
+.calendar__date:nth-child(4):before {
```
loralevitska, line 86: Use `scss` loop and variables for this logic

loralevitska (changes requested):
Keep up the good work! Some comments to fix:
- fix day count modifier 
image

PR #2636, opened 2023-08-22
Layout_calendar

VitaliiOkhotskyi:
- [DEMO LINK](https://VitaliiOkhotskyi.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,54 @@
+.calendar {
+  width: ($day-size + $gap) * $colums;
+
+  &__container {
+    display: flex;
+    flex-wrap: wrap;
+    gap: $gap;
+  }
+
+  &__day {
+    position: relative;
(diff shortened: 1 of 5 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  flex-wrap: wrap;
+
+  &__day {
+    position: relative;
+    box-sizing: border-box;
+
+    height: $day-size;
+    width: $day-size;
+
+    background-color: #eee;
```
loralevitska: It is a good practice to make all colors [variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

loralevitska (changes requested):
Almost done) Just need to fix this 
image
image

etojeDenys (approved):
lgtm

PR #2643, opened 2023-08-27
add task solution

wsadocha:
- [DEMO LINK](https://wsadocha.github.io/layout_calendar/)
- [TEST REPORT LINK](https://wsadocha.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,38 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar calendar--start-day_mon calendar--month-length_28">
-      <div class="calendar__day-1"></div>
-      <div class="calendar__day-2"></div>
-      <div class="calendar__day-3"></div>
-      <div class="calendar__day-4"></div>
-      <div class="calendar__day-5"></div>
-      <div class="calendar__day-6"></div>
-      <div class="calendar__day-7"></div>
-      <div class="calendar__day-8"></div>
-      <div class="calendar__day-9"></div>
-      <div class="calendar__day-10"></div>
-      <div class="calendar__day-11"></div>
-      <div class="calendar__day-12"></div>
-      <div class="calendar__day-13"></div>
-      <div class="calendar__day-14"></div>
-      <div class="calendar__day-15"></div>
-      <div class="calendar__day-16"></div>
-      <div class="calendar__day-17"></div>
-      <div class="calendar__day-18"></div>
-      <div class="calendar__day-19"></div>
-      <div class="calendar__day-20"></div>
-      <div class="calendar__day-21"></div>
-      <div class="calendar__day-22"></div>
-      <div class="calendar__day-23"></div>
-      <div class="calendar__day-24"></div>
-      <div class="calendar__day-25"></div>
-      <div class="calendar__day-26"></div>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
...
+      <div class="calendar__day-22"></div>
+      <div class="calendar__day-23"></div>
+      <div class="calendar__day-24"></div>
+      <div class="calendar__day-25"></div>
+      <div class="calendar__day-26"></div>
+      <div class="calendar__day-27"></div>
+      <div class="calendar__day-28"></div>
+      <div class="calendar__day-29"></div>
+      <div class="calendar__day-30"></div>
+      <div class="calendar__day-31"></div>
```
darokrk: The calendar includes 31 days, regardless of the month length. Consider adding a mechanism to set the number of days dynamically.

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+      justify-content: center;
+      align-items: center;
+      font-size: 30px;
+
+      &::before {
+        content: "#{$d}";
+      }
+
+      &:hover {
+        cursor: pointer;
```
darokrk: It's better to specify the 'transition' property outside of the 'hover' state. This way, the transition will apply both when the hover starts and when it ends. Move the 'transition-duration: 0.5s;' to the '&__day-#{$d}' selector.

src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day_mon calendar--month-length_28">
```
darokrk: You should use here `<ul>` tag with `<li>` to use more semantic HTML

```suggestion
    <ul class="calendar calendar--start-day_mon calendar--month-length_28">
      <li class="calendar__day-1"></li>
      <li class="calendar__day-2"></li>
      <li class="calendar__day-3"></li>
```

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
+$day-size: 100px;
```
darokrk: please move these all variables into newly created file variables.scss

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
+
+  &--month-length {
+    @for $length from 28 to 32 {
+      &_#{$length} :nth-child(n + #{$length + 1}) {
+        display: none;
+      }
+    }
+  }
+
+  @for $d from 1 to 32 {
```
darokrk: please change this variable name to be more descriptible `@for $day from 1 to 32 {`

darokrk (changes requested):
Layout looks good, left comments for the code improvements 🛠️

src/index.html
wsadocha: I don't know how to connect number of < li >'s in HTML, with the class name in which I set the month length. I tried to set only 28 < li >'s and then add some more with ::after pseudoclass, but it doesn't work with more than one day ;/ I don't know how to do that. Maybe I don't fully understand what I have to do, but I've tried my best :P

mykhalenych (approved):

PR #2662, opened 2023-09-04
Calendar

xhuhm:
- [DEMO LINK](https://xhuhm.github.io/layout_calendar/)
- [TEST REPORT LINK](https://xhuhm.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,38 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar calendar--first-day-tue calendar--month-length-29">
-      <div class="day"></div>
-      <div class="day"></div>
-      <div class="day"></div>
-      <div class="day"></div>
-      <div class="day"></div>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--first-day-tue calendar--month-length-29">
+      <div class="day"></div>
```
volodymyr-soltys97: Need to fix everywhere
```suggestion
      <div class="calendar__day"></div>
```

volodymyr-soltys97 (changes requested):
This part of the task is incomplete.
There are 8 columns on the page, which is too much because a week has 7 days and there should be 7 columns
image
image

anastasiiavorobiova (approved):
Awesome!

PR #2670, opened 2023-09-07
add calendar

DenysShurubura:
- [DEMO LINK](https://DenysShurubura.github.io/layout_calendar/)
- [TEST REPORT LINK](https://DenysShurubura.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,68 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+    </div>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -7,67 +7,37 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar">
+    <div class="calendar calendar--start-day--wed calendar--month-length--30">
       <div class="calendar__day"></div>
-
       <div class="calendar__day"></div>
-
       <div class="calendar__day"></div>
-
       <div class="calendar__day"></div>
-
       <div class="calendar__day"></div>
-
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+}
+
+@for $i from 28 through 31 {
+  .calendar.month-length-#{$i} .calendar__day:nth-child(n + #{$i + 1}) {
+    display: none;
+  }
+}
+
+@for $i from 1 through 31 {
+  .calendar .calendar__day:nth-child(n + #{$i})::before {
```
vlevik: good job! :)
but I have some comments
here you can write just .calendar__day:nth-child... (not to increase specificity of selector)

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+  padding: 20px;
+
+  &__day {
+    position: relative;
+    width: $day-width;
+    height: $day-heigth;
+    box-sizing: border-box;
+    border: 1px solid #000;
+    background-color: #eee;
+    font-size: 30px;
```
vlevik, line 45: you can also use here variables

vlevik (approved):

src/index.html
```diff
@@ -7,6 +7,68 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
```
polosanya: Modifiers for calendar block are absent. Would be great to add them initially to show the functionality 
image

src/index.html
```diff
@@ -7,6 +7,68 @@
...
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
+
+      <div class="calendar__day"></div>
```
polosanya: You can remove newlines between calendar's elements as they are empty

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+@for $i from 28 through 31 {
+  .calendar.month-length-#{$i} .calendar__day:nth-child(n + #{$i + 1}) {
```
polosanya: The modifier is not a separate block. It's a part of the calendar block
image

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+  $day-width: 100px;
+  $calendar-width: 7 * $day-width + 10px;
+  $month-length: 28,29,30,31;
+  $days: ("mon": 0,
+  "tue": $day-width + 1,
+  "wed": $day-width * 2 + 2,
+  "thu": $day-width * 3 + 3,
+  "fri": $day-width * 4 + 4,
+  "sat": $day-width * 5 + 5,
+  "sun": $day-width * 6 + 6);
```
polosanya: Better create a separate file to store all variables

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+
+  &__day {
+    position: relative;
+    width: $day-width;
+    height: $day-heigth;
+    box-sizing: border-box;
+    border: 1px solid #000;
+    background-color: #eee;
+    font-size: 30px;
+    transition-duration: 0.5s;
```
polosanya: Please specify the transition property. Also, would be great to make transition-duration a variable as it should be the same per page

polosanya (changes requested):
Nice! Let's improve the solution a bit)
Pay attention to the task description and checklist

src/styles/main.scss
```diff
@@ -1,3 +1,54 @@
...
+  }
+}
+
+@for $i from 1 through 31 {
+  .calendar__day:nth-child(n + #{$i})::before {
+    content: "#{$i}";
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
```
polosanya, line 21: Better leave only content here. You can set all other properties just under the `calendar__day` selector, no need to use loop for that. Btw, you can center the text using `text-align` and `line-height` properties

src/styles/variables.scss
```diff
@@ -0,0 +1,15 @@
+$calendar-day-bg-color: #eee;
+$calendar-day-border-color: #000;
+$calendar-day-hover-color: pink;
+$day-heigth: 100px;
+$day-width: 100px;
+$calendar-width: 7 * $day-width + 10px;
+$month-length: 28,29,30,31;
+$main-transition: all 0.5s;
```
polosanya, line 8: It is a good practice to specify the properties you want apply the transition to instead of `all`.
In that case, I think it's better to create a variable just for `transition-duration`

polosanya (approved):
Well done!

PR #2675, opened 2023-09-07
add task solution

Drach-Illya:
[DEMO LINK](https://Drach-Illya.github.io/layout_calendar/)


src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+  "thu" 3,
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
+
+$month-length:
+  28,
+  29,
+  30,
+  31;
```
adioskv: You are not using that `month-length` in your `main.scss` .

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
-body {
+$calendar-border: 20px 10px;
+$calendar-gaps: 1px;
+$day-size: 100px;
+$day-background: #eee;
+$day-background-hover: pink;
+$day-border-color: #000;
+$day-font: arial, helvetica, sans-serif;
```
adioskv: You are not reusing theese variables in code

but i can't say that this is a fault, variables are good 🙌.

src/styles/main.scss
adioskv:
```suggestion
```

adioskv (approved):
Good job! 🙌.  Just remove unnecessary variable, and this will be a perfect calendar 😉!

src/styles/main.scss
Drach-Illya: Lector Told us today that most projects will have all/most things like colors, fonts and sizes put into variable even if they are not reused. Additionally, once I start putting variables into a separate SCSS file, it will be easier to change them rather than delve into the file they are important for.

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
```
polosanya: Let's create a separate file for variables

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+  "fri" 4,
+  "sat" 5,
+  "sun" 6;
+
+body,
+p {
   margin: 0;
 }
+
+.calendar {
```
polosanya: image

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+  box-sizing: border-box;
+  display: flex;
+  flex-wrap: wrap;
+  gap: #{$calendar-gaps};
+  padding: #{$calendar-border};
+  width: 750px;
```
polosanya: How did you get this number? I guess it should be calculated, not hardcoded

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+p {
   margin: 0;
 }
+
+.calendar {
+  box-sizing: border-box;
+  display: flex;
+  flex-wrap: wrap;
+  gap: #{$calendar-gaps};
+  padding: #{$calendar-border};
```
polosanya:
```suggestion
  gap: $calendar-gaps;
  padding: $calendar-border;
```
You can use variables like this

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+    width: #{$day-size};
+    height: #{$day-size};
+
+    text-align: center;
+    line-height: 100px;
+    background-color: #{$day-background};
+    border: 1px solid #{$day-border-color};
+
+    cursor: pointer;
+    transition: transform 0.5s;
```
polosanya: What about background-color?
Also, let's create a variable for transition-duration as it should be the same per page

src/styles/main.scss
```diff
@@ -1,3 +1,73 @@
...
+    &:hover {
+      background-color: #{$day-background-hover};
+      transform: translate(0, -20px);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: "#{$i}";
+        font-family: $day-font;
+        font-size: 30px;
```
polosanya: Better add these properties just for the `calendar__date` selector. Leave only content here

polosanya (changes requested):

src/styles/main.scss
Drach-Illya: Can you please elaborate what you mean when you say: "What about background-color?"

polosanya: You change background-color on hover, but don't include it in transition

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,54 @@
...
+
+    text-align: center;
+    line-height: 100px;
+    font-family: $day-font;
+    font-size: 30px;
+    background-color: #{$day-background};
+    border: 1px solid #{$day-border-color};
+
+    cursor: pointer;
+    transition: all $day-transform-time;
```
polosanya, line 25: It is a good tone to specify the properties you want to apply transition to instead of `all`

src/styles/utilities/variables.scss
```diff
@@ -0,0 +1,18 @@
+$calendar-border-vert: 20px;
+$calendar-border-hor: 10px;
+$calendar-gaps: 1px;
+$day-size: 100px;
+$day-transform-time: 0.5s;
+$day-background: #eee;
+$day-background-hover: pink;
+$day-border-color: #000;
+$day-font: arial, helvetica, sans-serif;
```
polosanya, line 9:
```suggestion
$day-font: Arial, Helvetica, sans-serif;
```

polosanya (approved):
Great!

src/styles/utilities/variables.scss
Drach-Illya: I vaguely recall linter being upset with me when fonts where written in upper case, which prompted me to make them like this in the first place.

src/styles/blocks/calendar.scss
Drach-Illya: I will keep this in mind. Thank you.

PR #2676, opened 2023-09-07
made calendar

sintax1s:
[DEMO LINK](https://sintax1s.github.io/layout_calendar/)
[TEST REPORT LINK](https://sintax1s.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <main class="calendar calendar--start-mon calendar--month-length-31">
+    <main class="calendar calendar--start--mon calendar--month-length--31">
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <main class="calendar calendar--start-mon calendar--month-length-31">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <main class="calendar calendar--start-mon calendar--month-length-31">
```
Drach-Illya: [BEM] - Make sure to follow BEM naming convention for complex modifiers: 
block-name--modifier-name--modifier-value;

Single - counts as "space" in one part, double -- counts as a separation between different parts of a complex modifier.

Add an extra dash before "mon" and "31" here and in SCSS

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,5 @@
+$cell-size: 100px;
+$columns: 7;
+$padding: 10px;
```
Drach-Illya: [Important]

[STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.

Change padding to 20px 10px to add top paddings tall enough that, when you click a day on the top row, it does not move out of the top part of the screen.

Drach-Illya (changes requested):
Very good code. Here are some things that can be improved.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+  (#{$padding-horizontal} * 2) + (#{$gap} * #{$columns}));
+  padding: $padding-vertical $padding-horizontal;
+
+  @each $day , $value in $days {
+    &--start--#{$day} &__day:nth-child(1) {
+      margin-left: $cell-size * $value + ($gap * $value);
+    }
+  }
+
+  @for $i from 28 through 31 {
```
vasylmiskiv, line 15: Would be better if you would write the variable more correctly. For example `$day` as you did above in the prev loop

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,44 @@
...
+    height: $cell-size;
+    background-color: #eee;
+    cursor: pointer;
+
+    &:hover {
+      transform: translateY(-20px);
+      background-color: pink;
+    }
+
+    @for $i from 1 through 31 {
```
vasylmiskiv, line 38: The same according to comment above

vasylmiskiv (approved):
Everything looks good. But they are some moments

PR #2678, opened 2023-09-07
calendar task implementation

yuliabakun:
- [DEMO LINK](https://yulabakun.github.io/layout_calendar/)

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,52 @@
+.calendar {
+  display: flex;
+  flex-shrink: 0;
+  flex-wrap: wrap;
+  gap: $gap-between-columns;
+  padding: $calendar-padding;
+  width: ($column-size * $columns-in-row) +
+    ($gap-between-columns * $gaps-in-row) +
+    $calendar-padding;
+
+  @for $days from 28 through 31 {
+    &--month-length--#{$days} {
+      :nth-child(n + #{$days+1}) {
+        visibility: hidden;
+      }
+    }
+  }
+
+  @for $i from 1 through 31 {
+    :nth-child(#{$i})::before {
+      content: "#{$i}";
+      display: block;
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      line-height: $column-size;
+      text-align: center;
+    }
+  }
+
+  @each $weekday, $shift in $weekdays {
+    &--start-day--#{$weekday} {
+      :first-child {
+        margin-left: #{$shift};
+      }
+    }
+  }
+
+  &__day {
+    box-sizing: border-box;
+    height: $column-size;
+    width: $column-size;
+    border: 1px solid black;
+    background-color: #eee;
+    cursor: pointer;
+    transition: transform 0.5s;
+
+    &:hover {
+      background-color: lightpink;
+      transform: translateY(-20px);
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
+.calendar {
+  display: flex;
+  flex-shrink: 0;
```
PaninDmytro: Unnecessary property

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+  padding: $calendar-padding;
+  width: ($column-size * $columns-in-row) +
+    ($gap-between-columns * $gaps-in-row) +
+    $calendar-padding;
+
+  @for $days from 28 through 31 {
+    &--month-length--#{$days} {
+      :nth-child(n + #{$days+1}) {
+        visibility: hidden;
+      }
```
PaninDmytro: @for $days from 28 through 31 {
    &--month-length--#{$days}:nth-child(n + #{$days}) {
        visibility: hidden;
      }
      /*Less Syntax*/

PaninDmytro (approved):

PaninDmytro:
Well done!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+  gap: $gap-between-columns;
+  padding: $calendar-padding;
+  width: ($column-size * $columns-in-row) +
+    ($gap-between-columns * $gaps-in-row) +
+    $calendar-padding;
+
+  @for $days from 28 through 31 {
+    &--month-length--#{$days} {
+      :nth-child(n + #{$days+1}) {
+        visibility: hidden;
```
polosanya: Better use `display: none` instead of `visibility: hidden` to completely remove the day from the page

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @for $i from 1 through 31 {
+    :nth-child(#{$i})::before {
+      content: "#{$i}";
+      display: block;
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      line-height: $column-size;
+      text-align: center;
```
polosanya: Leave only content here. You can add all other properties using `calendar__day` selector, no need to use loop for that
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  &__day {
+    box-sizing: border-box;
+    height: $column-size;
+    width: $column-size;
+    border: 1px solid black;
+    background-color: #eee;
+    cursor: pointer;
+    transition: transform 0.5s;
```
polosanya: What about background-color? You change it on hover too.
Also, would be great to create a variable for transition-duration as it should be the same per page
image

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,7 @@
+$column-size: 100px;
+$columns-in-row: 7;
+$gap-between-columns: 1px;
+$gaps-in-row: $columns-in-row - 1;
+$calendar-padding: 10px;
+$weekdays:
+mon 0, tue 101px, wed 202px, thu 303px, fri 404px, sat 505px, sun 606px;
```
polosanya: How did you get these numbers? Would be great to calculate them instead of hardcoding

polosanya (changes requested):

src/styles/blocks/calendar.scss
polosanya: This property is really unnecessary

polosanya (approved):
Well done!

PR #2682, opened 2023-09-07
add task solution

adioskv:
- [DEMO LINK](https://adioskv.github.io/layout_calendar/)
- [TEST REPORT LINK](https://adioskv.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,2 +1,65 @@
...
+  }
+
+  &--start-day {
+    @each $day, $cell in $weekDays {
+      &--#{$day} .calendar__cell:first-child {
+        margin-left: $cell;
+      }
+    }
+  }
+
```
MVoutch: If you change the block size, you'll have problems with markup. That's why it's better to do this
image
image

MVoutch (changes requested):

src/styles/main.scss
MVoutch:
```
$weekDays:
  mon 0,
  tue 1,
  wed 2,
  thu 3,
  fri 4,
  sat 5,
  sun 6;
$blockSize: 100px;
$gap: 1px

.calendar {
  position: absolute;
  display: flex;
  flex-wrap: wrap;

  padding: $blockSize;
  width: 710px;
  gap: $gap;

  &--month-length {
    @for $day from 28 through 30 {
      &--#{$day} > .cell:nth-child(n+#{$day + 1}) {
        display: none;
      }
    }
  }

  &--start-day {
    @each $day, $cell in $weekDays {
      &--#{$day} > .cell:first-child {
        margin-left: $cell * $BlockSize + $gap * 7;
      }
    }
  }
}
```

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
-body {
-  margin: 0;
+$blockSize: 100px;
+$gapSize: 1px;
+$weekDays:
```
polosanya: Let's create a separate file for variables

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+  mon 0,
+  tue 1,
+  wed 2,
+  thu 3,
+  fri 4,
+  sat 5,
+  sun 6;
+
+.calendar {
```
polosanya: image

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+  position: absolute;
+  display: flex;
+  flex-wrap: wrap;
+
+  padding: $blockSize;
```
polosanya: It's up to you, but I'd recommend just to center the calendar instead of setting paddings

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+    box-sizing: border-box;
+    text-align: center;
+
+    width: $blockSize;
+    height: $blockSize;
+    line-height: $blockSize;
+
+    background-color: #eee;
+    border: 1px solid #000;
+    transition: transform 0.5s;
```
polosanya: What about background-color? You change it on hover too.
Also, would be great to create a variable for transition-duration
image

src/styles/main.scss
```diff
@@ -1,3 +1,66 @@
...
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i}) {
+        &::before {
+          content: "#{$i}";
+          font-family: Arial, Helvetica, sans-serif;
+          font-size: 30px;
```
polosanya: Leave only content here. You can add all other properties just with `calendar__cell` selector, no need to use loop for that
image

polosanya (changes requested):

src/styles/main.scss
adioskv: Can we use formula to change font-size depending on cell size,
  ignoring the README.md file ? If we are prepared for cell size change, it would be great
  to have a formula for font

StasSokolov1 (approved):
Great job🔥

PR #2687, opened 2023-09-07
add task Calendar solution

o-drozzdyk:
- [DEMO LINK](https://o-drozzdyk.github.io/layout_calendar/)


src/styles/blocks/day.scss
```diff
@@ -0,0 +1,18 @@
+.calendar {
+  &__day {
+    box-sizing: border-box;
+    width: $day-block-size;
+    height: $day-block-size;
+    border: 1px solid #000;
+
+    background-color: #eee;
+    text-align: center;
+    transition-duration: 0.5s;
```
dolphinnq: You can use just `transition`, `duration` in selector is not necessary here.

src/styles/blocks/day.scss
```diff
@@ -0,0 +1,18 @@
...
+
+    &:hover {
+      background-color: rgb(255, 192, 203);
```
dolphinnq: You can simply use `pink` here for `background-color`.

src/styles/blocks/startDay.scss
```diff
@@ -0,0 +1,11 @@
+.calendar {
+  $days: "mon", "tue", "wed", "thu", "fri", "sat", "sun";
```
dolphinnq: It is better to set certain values in a new line:
```
$days: 
  "mon",
  "tue"...
```

dolphinnq (approved):
Great job! Almost clean!

src/styles/blocks/day.scss
```diff
@@ -0,0 +1,18 @@
+.calendar {
+  &__day {
+    box-sizing: border-box;
```
vasylmiskiv: Better to move box-sizing to main file and apply the property to *

src/styles/blocks/daysNumbering.scss
```diff
@@ -0,0 +1,11 @@
+.calendar {
+  @for $number from 1 through 31 {
+    & :nth-child(#{$number})::before {
+      content: "#{$number}";
+      line-height: $day-block-size;
+      font-family: arial, Helvetica, sans-serif;
```
vasylmiskiv: Would be better to use `font-family` in body for instance

src/styles/blocks/daysNumbering.scss
```diff
@@ -0,0 +1,11 @@
+.calendar {
+  @for $number from 1 through 31 {
```
vasylmiskiv: dont use it in separate files write it only in `calendar.scss`. You shouldn't write the logic which nested in one class in separate file

src/styles/blocks/monthLength.scss
```diff
@@ -0,0 +1,9 @@
+.calendar {
+  @for $number from 28 through 31 {
```
vasylmiskiv: The same according to comment above about separate logic. Write it in calendar.scss file

src/styles/blocks/startDay.scss
vasylmiskiv: The same (comments above) and move `$day` to variables file and write it like so 
```
$days: (
  mon: 0,
  tue: 1,
  wed: 2,
  thu: 3,
  fri: 4,
  sat: 5,
  sun: 6
);
```

src/styles/blocks/startDay.scss
```diff
@@ -0,0 +1,11 @@
...
+  @each $day in $days {
+    $i: index($list: $days, $value: $day);
```
vasylmiskiv: remove this line and use it in `each` - `$day, $index in $days`

src/styles/blocks/startDay.scss
```diff
@@ -0,0 +1,11 @@
...
+    &--start-day--#{$day} {
+      & :first-child {
+        margin-left: $day-block-size * ($i - 1) + $gap-size * ($i - 1);
```
vasylmiskiv: now use `$index` instead of `$i`

src/styles/main.scss
```diff
@@ -1,3 +1,7 @@
-body {
-  margin: 0;
-}
+@import "./utils/variables";
+
+@import "./blocks/calendar";
+@import "./blocks/day";
```
vasylmiskiv: remove all unnecessary imports according to comments above

vasylmiskiv (changes requested):
Need to add some changes

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $day-block-size * 7 + $gap-size * 6 + $side-padding * 2;
```
polosanya, line 4: What are 7, 6, 2? Try to avoid magic numbers. Also, would be great to create a separate variable for calendar width and move all calculations there

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+  padding: 25px $side-padding $side-padding;
+  gap: $gap-size;
+
+  &__day {
+    box-sizing: border-box;
```
polosanya, line 9: You've already set it for all elements, no need to duplicate it here

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+      cursor: pointer;
+      transform: translatey(-20px);
+    }
+  }
+
+  @for $number from 1 through 31 {
+    & :nth-child(#{$number})::before {
+      content: "#{$number}";
+      line-height: $day-block-size;
+      color: #000;
```
polosanya, line 29: image
```suggestion
      content: "#{$number}";
```
Just move other properties under calendar__day selector

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+
+  &__day {
+    box-sizing: border-box;
+    width: $day-block-size;
+    height: $day-block-size;
+    border: 1px solid #000;
+
+    background-color: #eee;
+    text-align: center;
+    transition: 0.5s;
```
polosanya, line 16: Would be great to make transition-duration a variable as it should be the same for whole page.
Also, it's better to specify transition-properties
image

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+      content: "#{$number}";
+      line-height: $day-block-size;
+      color: #000;
+    }
+  }
+
+  @for $number from 28 through 31 {
+    &--month-length--#{$number} {
+      & :nth-child(n + #{$number + 1}) {
+        visibility: hidden;
```
polosanya, line 36: Not critical, but I think `display: none` is better to completely remove elements from page

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments)

PR #2692, opened 2023-09-08
made calendar more srtyle-correct

Ayogoharo:
- [DEMO LINK](https://ayogoharo.github.io/layout_calendar/)
- [CHECKLIST LINK](https://github.com/mate-academy/layout_calendar/blob/master/checklist.md)


src/index.html
```diff
@@ -6,7 +6,44 @@
...
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
-  </body>
+  <div class="calendar">
+    <div class="calendar__start-day-wed"></div>
+    <div class="calendar__month-length-29"></div>
+
+    <div class="calendar__sheet" >
```
HardCodWorker: Not sure that it's a good practice to split the modifiers into elements of the main block

You may reference check-list (p.2)

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
-body {
+$cell-size: 100px;
+$gap-size: 1px;
+$padding: 10px;
+$sheet-with: $cell-size * 7 + $gap-size * 6;
```
HardCodWorker: I would create a new scss file and put all the variables there. 

You may reference to checklist (p.5)

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
...
+
+  @each $name, $num in $days-in-week {
+    &__start-day-#{$name} {
+      ~ .calendar__sheet > :first-child {
+        margin-left: $cell-size * ($num - 1) + $gap-size * ($num - 1);
+      }
+    }
+  }
+
+  @for $n from 28 to 32 {
```
HardCodWorker: I would use a varialbe that would be more clear

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
...
+      color: #000;
+
+      font-family: $number-font;
+      font-size: $number-font-size;
+
+      &:hover {
+        cursor: pointer;
+        background-color: pink;
+        transform: translate(0, -20px);
+        transition: transform 0.5s;
```
HardCodWorker: Please reference to checklist(p.9)

HardCodWorker (changes requested):
Also as I can see there is something wrong with the sizing of the sheet as I could not detecte 10 px padding  that should be accroding to checklist(p.11) - [STYLES] - Don't be greedy, add some top paddings, so the top row don't go over top of the page on hover.

Seems like the main task had a poitn which said to implement the layout in the middle of the screen. 

image

Nevertheless, a very good one especially knowing that you managed to commit so early.

All the statements are just my opinion and I never mean to insult you =)

src/index.html
polosanya:
```suggestion
  <div class="calendar calendar--start-day--wed calendar--month-length--29">
```
Modifiers should be added to the block directly, they are not its elements
image

src/index.html
```diff
@@ -6,7 +6,44 @@
...
+      <div class="calendar__sheet-cell"></div>
+      <div class="calendar__sheet-cell"></div>
+      <div class="calendar__sheet-cell"></div>
+      <div class="calendar__sheet-cell"></div>
+      <div class="calendar__sheet-cell"></div>
+    </div>
+  </div>
+
+
 </html>
```
polosanya:
```suggestion
  </div>
</html>
```
Don't leave empty lines between parent and child elements. Also, body is absent

src/styles/main.scss
polosanya: ➕ , create a separate file for variables. Also, avoid magic numbers

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+$days-in-month: 28, 29, 30, 31;
+
+$number-font: "Arial", sans-serif;
+$number-font-size: ($cell-size * 0.3);
+$number-font-position: ($cell-size - $number-font-size) / 2;
+
+* {
   margin: 0;
+  padding: 0;
+  animation-duration: 10s;
```
polosanya: It is recommended to use * only for box-sizing as other properties can affect the performance.
Btw, why do you need animation-duration? You don't have any animations in your code

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+}
+
+.calendar {
+  padding: $padding;
+  display: flex;
+  justify-content: center;
+
+  &__start-day-wed,
+  &__month-length-29 {
+    display: none;
```
polosanya: Don't hardcode it here. You have a `for` for this feature

src/styles/main.scss
polosanya: ➕ , what is `n`? And why do you need 32? Calendar can have only 31 days max

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+      box-sizing: border-box;
+      height: $cell-size;
+      width: $cell-size;
+      background-color: #eee;
+      border: 1px solid #000;
+      color: #000;
+
+      font-family: $number-font;
+      font-size: $number-font-size;
+      transition: transform 0.5s;
```
polosanya: What about background-color? You change it on hover too.
Also, would be great to create a variable for transition duration
image

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+      }
+    }
+
+    @for $i from 1 to 32 {
+      &>:nth-child(#{$i})::before {
+        content: "#{$i}";
+        position: absolute;
+        top: $number-font-position;
+        display: block;
+        height: $number-font-size;
```
polosanya:
```suggestion
    @for $i from 1 to 31 {
      &>:nth-child(#{$i})::before {
        content: "#{$i}";
```
You can use class selector instead of `for` loop to add all other properties
image

src/styles/main.scss
```diff
@@ -1,3 +1,83 @@
...
+$number-font-size: ($cell-size * 0.3);
+$number-font-position: ($cell-size - $number-font-size) / 2;
+
+* {
   margin: 0;
+  padding: 0;
+  animation-duration: 10s;
+}
+
+.calendar {
```
polosanya: Create a separate file for this BEM block

polosanya (changes requested):
Good job! Let's improve your code a bit)
Pay attention to the checklist

src/styles/main.scss
Ayogoharo: cause i used `to` against `through` in for loop

src/styles/main.scss
Ayogoharo: I hardcoded it because we have only 28 -> 31 days in the month, no less no more

src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
-  </body>
+  <main class="calendar calendar--start-day--wed calendar--month-length--29">
+    <section class="calendar__sheet">
+      <div class="calendar__sheet-cell"></div>
```
polosanya: Don't make `main` a BEM block. Move all classes to the section

src/styles/utils/normalize.scss
```diff
@@ -0,0 +1,9 @@
+* {
+  animation-duration: 10s;
```
polosanya: Why do you need this property?

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+
+  @each $name, $num in $days-in-week {
+    &--start-day--#{$name} > {
+      .calendar__sheet > :first-child {
+        margin-left: $cell-size * ($num - 1) + $gap-size * ($num - 1);
+      }
+    }
+  }
+
+  @for $n from 28 to 32 {
```
polosanya:
```suggestion
  @for $n from 28 through 31 {
```
It will work the same. but looks better and easier to undestand

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+      justify-content: space-around;
+      box-sizing: border-box;
+      height: $cell-size;
+      width: $cell-size;
+      background-color: #eee;
+      border: 1px solid #000;
+      color: #000;
+      font-family: $number-font;
+      font-size: $number-font-size;
+      transition: transform $transitionDuration;
```
polosanya:
```suggestion
      transition: transform $transitionDuration, background-color $transformDuration;
```
Background color should change smoothly too

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+
+      &:hover {
+        cursor: pointer;
+        background-color: pink;
+        transform: translate(0, -20px);
+      }
+    }
+
+    @for $i from 1 to 32 {
```
polosanya: Same here

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+      &>:nth-child(#{$i})::before {
+        content: "#{$i}";
+        position: absolute;
+        top: $number-font-position;
+        height: $number-font-size;
```
polosanya: image
Don't use loop to set the properties that can be set just with the class selector. Store all properties in one place
```suggestion
        content: "#{$i}";
```

polosanya (changes requested):
Almost done!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+  }
+
+  &__cell {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    box-sizing: border-box;
```
StasSokolov1, line 35: you can remove this line

StasSokolov1 (approved):
Great job! Only one minor comment!

PR #2749, opened 2023-09-27
Add calendar solution

Sadeni-09:
- [DEMO LINK](https://Sadeni-09.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Sadeni-09.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -24,17 +24,23 @@
       background-color: #ffc0cb;
     }
 
-    @for $day-num from 1 through $days {
-      &:nth-child(#{$day-num})::before {
-        content: "#{$day-num}";
+    @for $day from 1 through $days {
+      &:nth-child(#{$day})::before {
+        content: "#{$day}";
       }
     }
   }
 
-  @each $day, $margin in $days-of-week {
-    --start-day--#{$day}
-    :first-child {
-      margin-left: #{$margin};
+  @each $day, $numOfDay in $days-of-week {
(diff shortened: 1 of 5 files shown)


lerastarynets (changes requested):
Nice start!
Screenshot at Sep 27 15-04-32
changing `calendar--start-day--` selector should change the layout of the calendar - you need to recalculate margins

Sadeni-09:
I didn't understand quite well your comment but I changed the way I have calculated calendar width, is it ok now?

src/styles/calendar.scss
```diff
@@ -0,0 +1,40 @@
...
+    @for $day-num from 1 through $days {
+      &:nth-child(#{$day-num})::before {
+        content: "#{$day-num}";
+      }
+    }
+  }
+
+  @each $day, $margin in $days-of-week {
+    --start-day--#{$day}
+    :first-child {
```
witflash:
```suggestion
    --start-day--#{$day} :first-child {
```
Could you try to use this syntax? Because I see your code doesn't include these generated selectors. So, I think there is a problem.

witflash (changes requested):
Check my comment, please.
Also, you have to implement `@for` cycle for the `calendar--month-length--#{$day-number}` selectors (from 28 to 31) to hide the last days if your calendar has less than 31.
For example:
```
@for $i from 28 through 31 {
```

anastasiia-tilikina (approved):
GJ 🔥

PR #2759, opened 2023-10-01
created a layout for month calendar

ebfidKate:
- [DEMO LINK](https://ebfidkate.github.io/layout_calendar/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,50 @@
+.calendar {
+  flex-wrap: wrap;
+  width: 7 * $squareSize + 6 * $gap;
+  gap: $gap;
+
+  $days: mon, tue, wed, thu, fri, sat, sun;
+
+  @each $day in $days {
+    &--start-day-#{$day} &__day:first-child {
+      $index: index($days, $day) - 1;
+      margin-left: $index * $squareSize + $index;
+    }
+  }
+
+  @for $length from 28 through 31 {
+    $firstHidden: $length + 1;
+
+    &--month-length-#{$length} &__day:nth-child(n + #{$firstHidden}) {
+      display: none;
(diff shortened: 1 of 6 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    @for $dayNum from 1 through 31 {
+      &:nth-child(#{$dayNum}) {
+        &::before {
+          content: "#{$dayNum}";
+          width: 100%;
+          margin-block: auto;
+          text-align: center;
+          font-size: 30px;
+        }
+      }
```
maxim2310, line 40:
```suggestion
        &::before {
          content: "#{$dayNum}";
          width: 100%;
          margin-block: auto;
          text-align: center;
          font-size: 30px;
        }
```

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
+.calendar {
+  flex-wrap: wrap;
+  width: 7 * $squareSize + 6 * $gap;
+  gap: $gap;
+
+  $days: mon, tue, wed, thu, fri, sat, sun;
```
maxim2310: please move variable to separate file

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

src/styles/calendar.scss
ebfidKate: > 
:nth-child pseudo-class is used here for targeting exact cell of calendar. If i'll just delete this part, i'll get 31 in each cell because the loop by itself targets every cell::before and on each step it rewrites previous labels. Could I leave it like that or should I change approach to labeling?

anastasiiavorobiova (approved):
Great work!

PR #2770, opened 2023-10-05
add task

vmisai:
[DEMO](https://vmisai.github.io/layout_calendar/)

I am not sure that I understand task in correct way(
But it works)


etojeDenys (approved):
good job

PR #2778, opened 2023-10-11
add task solution

TetianaBukhenko:


src/index.html
```diff
@@ -8,5 +8,38 @@
   </head>
   <body>
     <h1>Calendar</h1>
+    <div class="calendar">
```
Viktor-Kost: calendar should have modifiers (for amount of days and start day). Changing this modifiers we should see changed calendar

Viktor-Kost (changes requested):
Please provide DEMO link to check your solution

src/index.html
TetianaBukhenko: Oh, sure. Sorry. https://TetianaBukhenko.github.io/layout_calendar/

TetianaBukhenko: May I clarify (I probably got the task wrong), I should add modifiers to the main container "calendar", right? It is not possible to add the modifier to the first child, right?

VitaliyBondarenko1982 (approved):
Looks good

PR #2790, opened 2023-10-13
add task solution

pasha30061:
- [DEMO LINK](https://pasha30061.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,57 @@
+$height: 100px;
+$width: 100px;
+$padding: 10px;
+$gap: 1px;
+
 body {
   margin: 0;
+  padding: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: #{$gap};
+  padding: #{$padding};
+  width: $width * 7 + $gap * 6;
+
+  &__day {
+    height: #{$height};
+    width: #{$width};
+    background-color: #eee;
+    border: 1px solid black;
+    box-sizing: border-box;
+    position: relative;
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transition: transform 0.5s;
+      transform: translateY(-20px);
+    }
+  }
+
+  @for $date from 0 through 31 {
+    & > :nth-child(n+#{$date})::before {
+      content: "#{$date}";
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      position: absolute;
+      left: 50%;
+      top: 50%;
+      transform: translate(-50%, -50%);
+    }
+  }
+
+  $dayOfWeek: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+
+  @each $startDay, $numberWeek in $dayOfWeek {
+    &--start-day-#{$startDay} > :nth-child(1) {
+      margin-left: ($width+$gap)*$numberWeek;
+    }
+  }
 }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,7 +1,4 @@
-$height: 100px;
-$width: 100px;
-$padding: 10px;
-$gap: 1px;
+@import "variables.scss";
 
 body {
   margin: 0;
@@ -36,7 +33,7 @@ body {
   }
 
   @for $date from 0 through 31 {
-    & > :nth-child(n+#{$date})::before {
+    :nth-child(n+#{$date})::before {
       content: "#{$date}";
       font-family: Arial, Helvetica, sans-serif;
       font-size: 30px;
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
we should have the ability to change the calendar by changing the modifiers
https://github.com/mate-academy/layout_calendar/assets/70220667/ab03189d-a08a-4eee-afc3-394d8fa90f72

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+    width: #{$width};
+    background-color: #eee;
+    border: 1px solid black;
+    box-sizing: border-box;
+    position: relative;
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transition: transform 0.5s;
```
Esceype: Move this CSS prop to calendar__day selector as transform should be smooth to both states, when the mouse enters and when it leaves.

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+
+  $dayOfWeek: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+
+  @each $startDay, $numberWeek in $dayOfWeek {
+    &--start-day-#{$startDay} > :nth-child(1) {
+      margin-left: ($width+$gap)*$numberWeek;
+    }
+  }
+
+  @for $monthLenght from 0 through 31 {
```
Esceype: It should be from 28, because we haven`t month have less then 28 days or more then 31. Your modifier at the moment is broken because there is ability to set less then 28 days in month
image

Esceype (changes requested):
At all great job, only two moments that need to be fixed.

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+
+  $dayOfWeek: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
+
+  @each $startDay, $numberWeek in $dayOfWeek {
+    &--start-day-#{$startDay} > :nth-child(1) {
+      margin-left: ($width+$gap)*$numberWeek;
+    }
+  }
+
+  @for $monthLenght from 28 through 31 {
```
volodymyr-soltys97:
```suggestion
  @for $monthLength from 28 through 31 {
```

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+
+  @for $date from 0 through 31 {
+    & > :nth-child(n+#{$date})::before {
```
volodymyr-soltys97:
```suggestion
    :nth-child(n+#{$date})::before {
```

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
+$height: 100px;
+$width: 100px;
+$padding: 10px;
+$gap: 1px;
```
volodymyr-soltys97: Move it ti separate file

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+      font-family: Arial, Helvetica, sans-serif;
+      font-size: 30px;
+      position: absolute;
+      left: 50%;
+      top: 50%;
+      transform: translate(-50%, -50%);
+    }
+  }
+
+  $dayOfWeek: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
```
volodymyr-soltys97: Move it to separate file

volodymyr-soltys97 (changes requested):
Good job 🔥 
Let's improve your code

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+  gap: #{$gap};
+  padding: #{$padding};
+  width: $width * 7 + $gap * 6;
+
+  &__day {
+    height: #{$height};
+    width: #{$width};
+    background-color: #eee;
+    border: 1px solid black;
+    box-sizing: border-box;
```
OleziO: Move this property to the * selector and remove it from all other places.

src/styles/main.scss
```diff
@@ -1,3 +1,58 @@
...
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+  }
+
+  @for $date from 0 through 31 {
+    :nth-child(n+#{$date})::before {
+      content: "#{$date}";
+      font-family: Arial, Helvetica, sans-serif;
```
OleziO: Move to the html selector and remove it from all other places.

OleziO (changes requested):

pasha30061:
- [DEMO LINK](https://pasha30061.github.io/layout_calendar/)

src/styles/main.scss
```diff
@@ -1,3 +1,64 @@
...
+  flex-wrap: wrap;
+  gap: #{$gap};
+  padding: #{$padding};
+  width: $width * 7 + $gap * 6;
+
+  &__day {
+    height: #{$height};
+    width: #{$width};
+    border: 1px solid black;
+    box-sizing: border-box;
```
TarasHoliuk: this comment about  `box-sizing: border-box;` property, not about `background-color: #eee;`
now it looks suboptimal

switch mentioned properties places.
Some explanation: `box-sizing: border-box;` is the preferable value and if you work with such a setting it should be the same for all elements on the page to avoid confusing, so the `*` selector is correct place for it

TarasHoliuk (changes requested):
One more change - the previous comment was misunderstood. If you not sure about meaning it's always better to ask in the chat to make sure you get it right :slightly_smiling_face:

TarasHoliuk (approved):
:heavy_check_mark:

PR #2823, opened 2023-10-19
task solution

Fluebubble:
Я не розділяв main.scss на різні імпорти, т.к. не бачу в цьому сенсу, але якщо це треба, то розіб'ю.

- [DEMO LINK](https://Fluebubble.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Fluebubble.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+      &:nth-child(#{$day})::before {
+        content: "#{$day}";
+        left: 50%;
+        top: 50%;
+
+        font-size: 30px;
+        font-family: Arial, Helvetica, sans-serif;
+        position: absolute;
+
+        transform: translate(-50%, -50%);
```
And678: Move these common `::before` styles outside of `@for` (leave only `content`). Check out the compiled css files in `./dist`, these 6 lines will be repeated 31 times, greatly increasing the total css size

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  "wed": 2,
+  "thu": 3,
+  "fri": 4,
+  "sat": 5,
+  "sun": 6,
+);
+
+@function calendarWidth($boxWidth, $gap) {
+  @return (($boxWidth * 7) + ($gap * 6));
+}
```
And678: :fire:

And678 (changes requested):
Good work, but one improvement can be made

Fluebubble:
Done

src/styles/main.scss
```diff
@@ -1,3 +1,77 @@
...
+    }
+  }
+
+  &__day {
+    position: relative;
+    box-sizing: border-box;
+    border: 1px solid black;
+    @include boxSize($row-width);
+
+    transition: 500ms;
```
BudnikOleksii: Specify to what properties the transition should apply

src/styles/main.scss
```diff
@@ -1,3 +1,77 @@
...
+
+    &::before {
+      left: 50%;
+      top: 50%;
+
+      font-size: 30px;
+      font-family: Arial, Helvetica, sans-serif;
+      position: absolute;
+
+      transform: translate(-50%, -50%);
```
BudnikOleksii: Instead of `position: absolute;` use just flex to a parent element(`&__day`)

src/styles/main.scss
```diff
@@ -1,3 +1,77 @@
...
+    }
+
+    @for $day from 1 through 31 {
+      &:nth-child(#{$day}):hover {
+        background-color: purple;
+        color: wheat;
+
+        transform: translateY(-20px);
+      }
```
BudnikOleksii: This styles are the same for all day elements so move them outside the loop, apply just once to `:hover` selector

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions. Also, from the task description:
- cursor should become pointer

src/styles/main.scss
Fluebubble: logical, thanks

src/styles/main.scss
Fluebubble: I didn’t specify it because I want it to apply to every property

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+  }
+
+  @for $i from 28 through 31 {
+    &--month-length-#{$i} :nth-child(n + #{$i + 1}) {
+      display: none;
+    }
+  }
+
+  &__day {
+    // position: relative;
```
roman-mirzoian: Do not leave unnecessary comments.

src/styles/main.scss
roman-mirzoian: @Fluebubble If you set transitions for all properties, you capture those that you will not use, which will once again overload your styles.
This won't be noticeable on simple tasks, but on large ones it will be very noticeable, so specifying specific properties is good practice for specifying styles

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
+$row-width: 100px;
```
roman-mirzoian: @Fluebubble  I read your comment about putting everything in different files, but it is a very good practice to put variables and functions separately from all rules, this will increase the readability and maintainability of the code. Therefore, it would be cool to take it out in a separate file 😎

src/styles/main.scss
```diff
@@ -1,3 +1,74 @@
...
+@mixin boxSize($value) {
+  height: #{$value};
+  width: #{$value};
+}
+
 body {
   margin: 0;
 }
+
+.calendar {
```
roman-mirzoian: If you add to this class `margin: 0 auto;` then the calendar is aligned horizontally in the center.

roman-mirzoian (changes requested):
Almost done!

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  &__day {
+    display: flex;
+    box-sizing: border-box;
+    border: 1px solid black;
+    @include boxSize($row-width);
+
+    transition: transform 500ms;
```
BudnikOleksii, line 27: Still not fixed 👀, Roman described why it's important

BudnikOleksii (approved):
Good work 👍

src/styles/blocks/calendar.scss
Fluebubble: sure? 
    transition: **_transform_** 500ms;
"transform" added.

BudnikOleksii: > sure? transition: **_transform_** 500ms; "transform" added.

On hover, you change the not only transform, but also the color and background color, you should specify them all so the transition will be smooth 

```
transition: transform 500ms, color 500ms, background-color 500ms;
```

PR #2829, opened 2023-10-20
add task solution

ilyapgitt:
- [DEMO LINK](https://ilyapgitt.github.io/layout_calendar/)

- [x] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,13 +1,13 @@
 .calendar {
   display: flex;
-  max-width: 730px;
+  max-width: $max-width;
   padding: 10px;
   flex-wrap: wrap;
-  gap: 1px;
+  gap: $calendar-gap;
   margin: 0 auto;
   font-family: Arial, Helvetica, sans-serif;
 
-  @each $day-name, $day in mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6 {
+  @each $day-name, $day in $weekdays {
     &--start-day--#{$day-name} {
       .calendar__day:nth-child(1) {
         margin-left: 100px * $day + $day;
@@ -17,35 +17,34 @@
 
   @for $day from 1 to 31 {
     &--days-#{$day} {
-      .calendar__day:nth-child(n + #{$day + 1}) {
+      &__day:nth-child(n + #{$day + 1}) {
         display: none;
       }
     }
   }
 
   &__day {
-    height: 100px;
-    width: 100px;
+    height: $height-day;
+    width: $width-day;
     background-color: #eee;
     border: 1px solid black;
     display: flex;
     font-size: 30px;
+    box-sizing: border-box;
     align-items: center;
     justify-content: center;
-    transition: transform 500ms;
+    transition: transform 500ms, background-color 500ms;
   }
 
   &__day:hover {
     cursor: pointer;
     transform: translateY(-20px);
     background-color: palevioletred;
-    animation-duration: 500ms;
   }
 
   @for $date from 1 through 31 {
     &__day:nth-child(#{$date})::before {
       content: "#{$date}";
-      display: block;
     }
   }
 }
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,13 +1,13 @@
-$weekdays: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
-
 .calendar {
   display: flex;
   max-width: 730px;
-  margin: 10px;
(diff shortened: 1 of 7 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
+$weekdays: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
```
roman-mirzoian: Better to keep variables in separate files.

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.calendar {
```
roman-mirzoian, line 1: If you add to this class `margin: 0 auto;` then the calendar is aligned horizontally in the center.

roman-mirzoian (changes requested):
Almost done!
According to the task, the text style should be in `Arial`:
image

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  display: flex;
+  max-width: 730px;
```
BudnikOleksii: If we change the calendar width for example to 200, your calendar will be broken.
max-width in this case should depends on `calendar__day` width

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  gap: 1px;
+  margin: 0 auto;
+  font-family: Arial, Helvetica, sans-serif;
+
+  @each $day-name, $day in mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6 {
+    &--start-day--#{$day-name} {
+      .calendar__day:nth-child(1) {
+        margin-left: 100px * $day + $day;
+      }
+    }
```
BudnikOleksii: Why did you move `mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6` from variable?
Move it back and as Roman mentioned move it to a separate file.
`100px` looks like it is a `calendar__day` width so it should be stored in a variable too

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @for $day from 1 to 31 {
+    &--days-#{$day} {
+      .calendar__day:nth-child(n + #{$day + 1}) {
```
BudnikOleksii: Use nesting

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  &__day {
+    height: 100px;
+    width: 100px;
+    background-color: #eee;
+    border: 1px solid black;
+    display: flex;
+    font-size: 30px;
+    align-items: center;
+    justify-content: center;
+    transition: transform 500ms;
```
BudnikOleksii:
```suggestion
    transition: transform 500ms, background-color 500ms;
```
You can also specify transition to `background-color` and it changes smoothly too

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  &__day:hover {
+    cursor: pointer;
+    transform: translateY(-20px);
+    background-color: palevioletred;
+    animation-duration: 500ms;
```
BudnikOleksii:
```suggestion
```
You already specified `duration` for transition property

src/styles/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @for $date from 1 through 31 {
+    &__day:nth-child(#{$date})::before {
+      content: "#{$date}";
+      display: block;
```
BudnikOleksii: `display: block;` is unnecessary

src/styles/main.scss
```diff
@@ -1,3 +1,3 @@
-body {
-  margin: 0;
-}
+@import "reset";
+@import "calendar";
+@import "variables.scss";
```
BudnikOleksii: I don't see any `variables.scss`, create it and move your variables to it

src/styles/reset.scss
```diff
@@ -0,0 +1,8 @@
+body,
+div {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  vertical-align: baseline;
```
BudnikOleksii: `div` doesn't have these properties by default, so it doesn't make sense

BudnikOleksii (changes requested):
Good start 👍 
Let's improve your solution. I left some suggestions

BudnikOleksii (approved):
Nicely done 🔥

PR #2832, opened 2023-10-20
calendar task solution

VeDono:
- [DEMO LINK](https://VeDono.github.io/layout_calendar/)

### DIFF
--- a/src/styles/components/calendar.scss
+++ b/src/styles/components/calendar.scss
@@ -1,0 +1,51 @@
+.calendar {
+  width: $calendar-width;
+  margin: 0 auto;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  transition: 0.2s ease-in-out;
+
+  // adding correct margin-left to needed ceil of day
+  @each $day, $dayNum in $daysWeek {
+    &--start-day-#{$day} {
+      .calendar__day:nth-child(1) {
+        margin-left: $ceil-width * $dayNum + $dayNum;
+      }
+    }
+  }
+
+  // disable ceils below count of day in month
+  @for $day from 1 through $daysCount {
+    &--month-length-#{$day} {
+      .calendar__day:nth-child(n + #{$day + 1}) {
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    @include square;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #fff;
+    font-size: 31px;
+    font-family: Arial, sans-serif;
+    transition: background-color, transform 0.5s ease-in-out;
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transform: translateY(-20px);
+    }
+        
+    // adding number of day in to ceils
+    @for $day from 1 through $daysCount {
+      &:nth-child(#{$day}):before {
+        content: "#{$day}";
+      }
+    }
+ 
+  }
+}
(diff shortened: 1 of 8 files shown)


src/styles/components/calendar.scss
```diff
@@ -0,0 +1,51 @@
+.calendar {
+  width: $calendar-width;
+  margin: 0 auto;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  transition: 0.2s ease-in-out;
+
+  // adding correct margin-left to needed ceil of day
```
BudnikOleksii: - Your calendar doesn't have any effects, only a `calendar __day` 😉 
- get rid of all comments

src/styles/components/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+  @each $day, $dayNum in $daysWeek {
+    &--start-day-#{$day} {
+      .calendar__day:nth-child(1) {
```
BudnikOleksii: Use nesting for selectors, fix it in all places

src/styles/components/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__day {
+    @include square;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #fff;
+    font-size: 31px;
+    font-family: Arial, sans-serif;
+    transition: background-color, transform 0.5s ease-in-out;
```
BudnikOleksii:
```suggestion
    transition: background-color 0.5s ease-in-out, transform 0.5s ease-in-out;
```
In your case `background-color` doesn't change smoothly, you should specify the duration too

src/styles/components/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    }
+        
+    // adding number of day in to ceils
+    @for $day from 1 through $daysCount {
+      &:nth-child(#{$day}):before {
+        content: "#{$day}";
+      }
+    }
+ 
+  }
```
BudnikOleksii: Remove blank line

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin square {
+  width: $ceil-width;
+  height: $ceil-height;
+  border: 1px solid black;
```
BudnikOleksii: `square` could be without a border too, so it's better to remove the border from mixin

BudnikOleksii (changes requested):
Looks good 🚀 
Just a few suggestions to improve your solution

src/styles/components/calendar.scss
VeDono: I fixed:

- rid of transition on 7 line and fix transition on 35 line.
- seems to fix nesting in 'for' & 'each'. (i not good with 'for' and 'each')
- rid of comments. (it's sad cause maybe in future i'm returns to this sample)
- fix blank line. (don't understand why linter catch error without this space, but now that fixed)

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin square {
+  width: $ceil-width;
+  height: $ceil-height;
+}
```
BudnikOleksii, line 4: It's better to accept width as a parameter, so in this case, you can create different squares

BudnikOleksii (approved):
Nicely done 🚀

PR #2847, opened 2023-10-28
Calendar HW

Denys209818:


Denys209818:
[DEMO LINK](https://denys209818.github.io/layout_calendar/)
[TEST REPORT LINK](https://denys209818.github.io/layout_calendar/report/html_report/)

volodymyr-soltys97 (changes requested):
Good job 👍 
1. This part of the task is incomplete
image

src/styles/main.scss
```diff
@@ -19,6 +19,13 @@ body {
   font-family: Arial, Helvetica, sans-serif;
 }
 
+.centered {
```
TarasHoliuk: It's still not centered
How does `position: absolute` work?

image

Hint: you don't need position: absolute here, you may use margin. Also, additional wrapper is redundant - add styles to body (but using class selector)

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+  @each $item, $left in $marginFromLeftSide {
+    &--start-day--#{$item} :first-child {
+      margin-left: calc(#{$left} * 101px);
+    }
+  }
+
+  @for $item from 28 through 31 {
+    &--month-length--#{$item} {
+      & :nth-child(-n + #{$item}) {
+        display: block;
```
TarasHoliuk: if you do the same vice-versa (display: none for items you wanna hide and display: block by default) less amount of CSS code will be generated from SCSS

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+      bottom: 0;
+      display: none;
+      transition: all 0.5s;
+      transition-timing-function: cubic-bezier(0, 0.48, 0.31, 0.93);
+      box-sizing: border-box;
+      width: $sideOfSquare;
+      height: $sideOfSquare;
+      border: 1px solid black;
+      background-color: #eee;
+      font-size: 30px;
```
TarasHoliuk: all these properties are the same for every item, move them out of `for` loop, to avoid repeating them. Just define them for `&__day` as usual

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+        position: absolute;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
+      }
+    }
+  }
+
+  &__day:hover {
+    bottom: 20px;
```
TarasHoliuk: it's more efficient to use transform for it

also, add a transition (but not inside `:hover` selector - directly for item, to make it work in both ways)

TarasHoliuk (changes requested):

src/styles/main.scss
TarasHoliuk: This comment still hasn't been fixed yet (the calendar is still not centered) - please fix it. If you have any questions - feel free to ask in the chat

TarasHoliuk (changes requested):

src/styles/main.scss
```diff
@@ -1,3 +1,80 @@
...
+);
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.page {
+  width: 100%;
+  height: 100%;
```
anastasiiavorobiova, line 24: Screenshot 2023-10-29 at 15 43 48
Add these styles to the page to center your calendar vertically

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/styles/main.scss
Denys209818: If I used display flex, would it be better to use justify-content and align-items?

VitaliyBondarenko1982 (approved):
Well done

PR #2860, opened 2023-11-02
add task solution

Val-Bryks:
- [DEMO LINK](https://Val-Bryks.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,10 @@
+@import "./utils/variables";
+@import "./blocks/calendar";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
 }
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
we must have ability to change the nr of cells by changing the modifier number, let's fix that
https://github.com/mate-academy/layout_calendar/assets/70220667/be3b3b14-b42c-4e2f-ae57-718f2c986b44

src/styles/main.scss
```diff
@@ -1,3 +1,10 @@
+@import "./utils/variables";
+@import "./blocks/calendar";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
 }
```
anastasiiavorobiova, line 10:
```suggestion

```
It's better to move your general styles to a separate file to keep your main file clean and clear

anastasiiavorobiova (approved):
Awesome work!

PR #2862, opened 2023-11-03
replaced loop @each by loop @for

Poviakalo:
- [DEMO LINK](https://Poviakalo.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Poviakalo.github.io/layout_calendar/report/html_report/)
- [ ] Changing 'month-lengh' and 'start-day' modifier in the code element
- [ ] Each day has no modifiers, only class (eg. calendar__day)
- [ ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--fri calendar--month-length--29">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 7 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
When I chose 20 in the modifier month length I get this problem
image

Will be better if you fix if

Poviakalo:
I'm disagree. Becauset there are from 28 to 31 days in the month. 

- Add a modifier `month-length` for the `calendar` with values 28, 29, 30 and 31 (use `@for`) - this is our task

src/index.html
```diff
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day--fri calendar--month-length--29">
```
Esceype, line 10:
```suggestion
    <div class="calendar calendar--start-day--fri calendar--month-length--29">
```
It should work only for 28-31 modifiers
Add a modifier month-length for the calendar with values 28, 29, 30 and 31 (use @for) 
But it should show all the days when someone put some number less than 28 or greater than 31
So you need to change logic of this modifier(it should set display: none) and by default, calendar__day should be display: block.

Esceype (changes requested):

andriy-shymkiv (approved):

PR #2914, opened 2023-11-27
'add task solution'

DenLysiak:
- [DEMO LINK](https://DenLysiak.github.io/layout_calendar/)
- [TEST REPORT LINK](https://DenLysiak.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -8,7 +8,7 @@
 
   @each $day, $margin in $days {
     &--start-day--#{$day} {
-      .calendar__day:first-of-type {
+      .calendar__day:first-child {
         margin-left: $margin;
       }
     }
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -1,0 +1,50 @@
+
+.calendar {
+  width: $calendar--size;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $day-block--gap;
+  padding: 10px;
+
+  @each $day, $margin in $days {
+    &--start-day--#{$day} {
+      .calendar__day:first-of-type {
+        margin-left: $margin;
+      }
+    }
+  }
+
+  @for $month-length from 28 through 31 {
+    &--month-length-#{$month-length} {
+      :nth-child(n + #{$month-length + 1}) {
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+
+    width: $day-block--size;
+    height: $day-block--size;
+
+    background-color: #eee;
+    border: 1px black solid;
+    transition: all 0.5s;
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(1n + #{$i})::before {
+        content: "#{$i}";
+      }
+    }
+  }
+}
--- a/src/styles/calendar.scss
+++ b/src/styles/calendar.scss
@@ -8,7 +8,7 @@
 
   @each $day, $margin in $days {
     &--start-day--#{$day} {
-      .calendar__day:first-child {
+      &__day:first-child {
         margin-left: $margin;
       }
     }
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,60 @@
...
+.calendar {
+  width: $calendar--size;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $day-block--gap;
+  padding: 10px;
+
+  @each $day, $margin in $days {
+    &--start-day--#{$day} {
+      div:first-child {
```
TarasHoliuk: it's better to use class selector

src/styles/main.scss
```diff
@@ -1,3 +1,60 @@
...
+  transition: transform 0.5s;
+
+  &:hover {
+    background-color: pink;
+    cursor: pointer;
+    transform: translateY(-20px);
+  }
+}
+
+@for $i from 1 through 31 {
```
TarasHoliuk: you may write it inside `.calendar__day` selector, using SCSS nesting with `&`

the same for `.calendar__day` - you may write it inside `.calendar` using nesting `&__day`

src/styles/main.scss
```diff
@@ -1,3 +1,60 @@
...
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+
+  width: $day-block--size;
+  height: $day-block--size;
+
+  background-color: #eee;
+  border: 1px black solid;
+  transition: transform 0.5s;
```
TarasHoliuk: background-color also changes, would be nice to add transition for it too

src/styles/main.scss
```diff
@@ -1,3 +1,60 @@
...
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+
+  display: flex;
+  justify-content: center;
+}
+
+.calendar {
```
TarasHoliuk: since it's BEM block let's move it to a separate file to stick the rules

Also, add some margin top, or padding for the body, etc, to prevent this:
image

TarasHoliuk (changes requested):
Well done :+1: 
The main logic of this task is written perfect, but let's improve a few small things)

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+.calendar {
+  width: $calendar--size;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $day-block--gap;
+  padding: 10px;
+
+  @each $day, $margin in $days {
+    &--start-day--#{$day} {
+      .calendar__day:first-child {
```
anastasiia-tilikina:
```suggestion
      &__day:first-child {
```

src/styles/calendar.scss
```diff
@@ -0,0 +1,50 @@
...
+        margin-left: $margin;
+      }
+    }
+  }
+
+  @for $month-length from 28 through 31 {
+    &--month-length-#{$month-length} {
+      :nth-child(n + #{$month-length + 1}) {
```
anastasiia-tilikina:
```suggestion
      & :nth-child(n + #{$month-length + 1}) {
```

anastasiia-tilikina (changes requested):
Almost done!

just a few suggestions. and pay attention that your month-length modifier doesn't work now. I guess the problem is in my second comment

src/styles/calendar.scss
DenLysiak: I have tried to use `&__day` but it is not working, and I don't understand why.

src/styles/calendar.scss
DenLysiak: On my computer, in vsCode everything is working good, while adding '&' breaks the loop, and it stops working. I will add changes anyway.

etojeDenys (changes requested):
image
image
looks like this. modifier doesn't work(should be 30 days when modifier calendar--month-length-30)


feel free to ask for some help in the chat

DenLysiak:
> image image looks like this. modifier doesn't work(should be 30 days when modifier calendar--month-length-30)
> feel free to ask for some help in the chat

that is so strange I have chacked demo link a few times and it is working, i dont realy understand what is going on

anastasiiavorobiova (approved):
Awesome work!

PR #2925, opened 2023-11-30
add solution

ssaviak:
- [DEMO LINK](https://ssaviak.github.io/layout_calendar/)
- [TEST REPORT LINK](https://ssaviak.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,9 +1,16 @@
-@import './utils/variables.scss';
+@import './utils/variables';
+
+* {
+  box-sizing: border-box;
+}
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
+}
 
 body {
   margin: 0;
-  font-family: Arial, Helvetica, sans-serif;
-  font-size: 30px;
 }
 
 .calendar {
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,51 @@
+@import './utils/variables.scss';
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
```
SerhiiKirik, line 9: move it to the `html` selector
```suggestion
```

src/styles/main.scss
```diff
@@ -1,3 +1,51 @@
...
+    justify-content: center;
+
+    height: $height-day-box;
+    width: $width-day-box;
+    border: 1px solid black;
+
+    &:hover {
+      cursor: pointer;
+      background-color: pink;
+      transition: $transition-during;
```
SerhiiKirik: Move to the `__day`, check, if you unhover - it would be without duration

SerhiiKirik: don't use it like that, because it's the same as transition: all 0.5s
It's too hard for the browser
Check this

Create a variable for transition duration
For example
```suggestion
    transition: color $transitionDuration, transform $transitionDuration;
```

src/styles/main.scss
```diff
@@ -1,3 +1,51 @@
...
+
+.calendar {
+  display: flex;
+  width: $width-max-calendar;
+  flex-wrap: wrap;
+  gap: $gap-between-days;
+  padding: $padding-between-days;
+
+  &__day {
+    box-sizing: border-box;
```
SerhiiKirik: move it to the `*` selector
```suggestion
```

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution & learn something new  🌍

dinaminahi (approved):
Good job 👍

PR #2932, opened 2023-12-01
add solution

ValeriaMyronets:
- [DEMO LINK](https://ValeriaMyronets.github.io/layout_calendar/)
- [TEST REPORT LINK](https://ValeriaMyronets.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <article class="calendar calendar--start--sun calendar--length--29">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+    </article>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <article class="calendar calendar--start--sun calendar--length--29">
+    <article class="calendar calendar--start--wed calendar--length--29">
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -42,3 +42,4 @@
     </article>
   </body>
 </html>
+
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,70 @@
+$size-cell: 100px;
+$general-font-size: 30px;
+$gap: 1px;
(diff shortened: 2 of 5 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,70 @@
...
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: $general-font-size;
+}
+
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
olya-shyian, line 11: its not good to write flex styles on body

src/styles/main.scss
```diff
@@ -1,3 +1,70 @@
...
+  }
+}
+
+@for $i from 1 through 31 {
+  .calendar .calendar__day:nth-child(#{$i})::before {
+    content: '#{$i}';
+  }
+}
+
+$start-day: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
```
olya-shyian: put this variable to all your varibles, try to save them all in one place

src/styles/main.scss
```diff
@@ -1,3 +1,70 @@
+$size-cell: 100px;
+$general-font-size: 30px;
+$gap: 1px;
```
olya-shyian: please create a separate file for your variables

src/styles/main.scss
```diff
@@ -1,3 +1,70 @@
...
+  }
+ }
+
+ @for $i from 28 through 30 {
+  @for $j from $i + 1 through 31 {
+    .calendar--length--#{$i} .calendar__day:nth-child(#{$j}) {
+      visibility: hidden;
+    }
+  }
 }
```
olya-shyian, line 69: good practice is to leave 1 line at the end of the file (please do this fix in each your file)

olya-shyian (changes requested):
Hats off to you! Great work🌸

src/styles/main.scss
ValeriaMyronets: But if I remove the display, the calendar isn't displayed in the center of the page, so I need to wrap it in another div. Maybe you have some other suggestions?

olya-shyian: yes, it's a possible option; but in our case I think if you remove this flex styles, your calendar still will be in center) let's try it

src/styles/main.scss
```diff
@@ -1,3 +1,71 @@
...
```
olya-shyian, line 69: please add a line at the end of the file

src/styles/var.scss
```diff
@@ -0,0 +1,4 @@
+$size-cell: 100px;
+$general-font-size: 30px;
+$gap: 1px;
+$start-day: mon 0, tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
```
olya-shyian: please add a line at the end of the file

src/styles/var.scss
```diff
@@ -0,0 +1,4 @@
+$size-cell: 100px;
```
olya-shyian: please update the name of your file; `var.scss` is not a good option, it should be more clear name like `variables.scss`

olya-shyian (changes requested):
well done!🔥

src/styles/main.scss
ValeriaMyronets: I tried to remove the flex, but the calendar moves up, that's why I asked you)

olya-shyian (changes requested):
Well done! ☀️
If in your code replace the modifier to another value, for example "calendar--length--31 calendar--start--mon" then the calendar will be broken, please fix this moment

https://www.loom.com/share/d0e6db0cd54441f6b8366d10121d03d1

src/index.html
```diff
@@ -7,6 +7,39 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <article class="calendar calendar--start--wed calendar--length--29">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
```
SerhiiKirik, line 12: would be better to do it with `ui` & `li` tags

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+}
+
+.calendar {
+  display: flex;
+  align-items: center;
+
+  flex-wrap: wrap;
+  gap: $gap;
+
+  width: $size-cell * 7 + $gap * 6;
```
SerhiiKirik, line 24: what is 7, 6? Would be better to move it to the variable

src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+    background-color: #eee;
+  }
+
+  &__day:hover {
+    cursor: pointer;
+    background-color: pink;
+
+    transform: translate(0, -20px);
+    transition-property: transform;
+    transition-duration: .5s;
```
SerhiiKirik, line 47: move it to the `__day` bc you don't have transition on unhover
```suggestion
    background-color: pink;

    transform: translate(0, -20px);
```

SerhiiKirik (approved):
Great 🔥

PR #2938, opened 2023-12-01
add task solution

YaroslavHor:
- [DEMO LINK](https://YaroslavHor.github.io/layout_calendar/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
image
looks broken when i set Sunday as start day, try to fix that

volodymyr-soltys97 (changes requested):
1. When I set the mon in the modifier then visually the month does not start from Monday
image

2. Still not fixed from the previous review
image

3. This part of the task is incomplete
image
image

etojeDenys (approved):
good job

PR #2945, opened 2023-12-02
Develop

Ar4ik12:
- [DEMO LINK](https://Ar4ik12.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Ar4ik12.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -26,7 +26,7 @@ body {
 .calendar {
   margin: 0 auto;
   padding: $padding-size;
-  width: (($day-size + $gap-size)*7)+$padding-size*2;
+  width: (($day-size + $gap-size) * 7) + $padding-size * 2;
   display: flex;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,69 @@
...
+
+    @for $i from 1 through 31 {
+      &:nth-child(n + #{$i})::after {
+        content: '#{$i}';
+      }
+    }
+
+
+    &:hover {
+      transition: transform 0.5s;
```
ghost, line 51: The transition property should be set on the element itself, not on the hover state. This will make the transition smooth in both directions (hover in and hover out).

ghost (changes requested):
Decent work! See the comment and also make the calendar, in the middle of the screen as vertically as horizontally!

PR #2952, opened 2023-12-03
Develop

YevhenOstrovskiy:
- [DEMO LINK](https://YevhenOstrovskiy.github.io/layout_calendar/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -4,9 +4,12 @@
   box-sizing: border-box;
 }
 
+html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
 body {
   margin: 0;
-  font-family: Arial, Helvetica, sans-serif;
 }
 
 .calendar {
@@ -17,6 +20,7 @@ body {
   width: $calendar-width;
   gap: $gap;
   padding-inline: $calendar-padding;
+  padding-top: 30px;
 
   &__day {
     display: flex;
@@ -30,7 +34,7 @@ body {
     background-color: $color-of-background;
     border: 1px solid $color-of-border;
     cursor: pointer;
-    transition: 0.5s;
+    transition: transform $transitionDuration;
 
     &:hover {
       cursor: pointer;
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,4 +1,4 @@
-@import './variables.scss';
+@import './variables';
 
 body {
   margin: 0;
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -47,7 +47,7 @@ body {
 
   @each $start-day, $i in $start-days {
     &--start-day--#{$start-day} &__day:first-child {
-      margin-left: ($box-size + $gap) *$i;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: left;
+  width: $calendar-width;
+  padding: 10px;
+  gap: $gap;
+
+
+  &__day {
+    box-sizing: border-box;
```
SerhiiKirik: Remove it to the `*` selector
```suggestion
```

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.calendar {
+  box-sizing: border-box ;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: left;
```
SerhiiKirik, line 18: let's align with center

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
+@import './variables';
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.calendar {
+  box-sizing: border-box ;
```
SerhiiKirik:
```suggestion
```

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
+@import './variables';
+
 body {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
```
SerhiiKirik: move to the `html`
```suggestion
```

src/styles/main.scss
```diff
@@ -1,3 +1,55 @@
...
+    display: flex;
+    width: $box-size;
+    height: $box-size;
+    position: relative;
+    font-size: 30px;
+    align-items: center;
+    justify-content: center;
+    background-color: $color-of-background;
+    border: 1px solid $color-of-border;
+    transition: 0.5s;
```
SerhiiKirik: don't use it like that, because it's the same as transition: all 0.3s
It's too hard for the browser
Check this

Create a variable for transition duration
For example
```suggestion
    transition:  transform $transitionDuration;
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some problems 🧑‍💻 
https://www.loom.com/share/d0e6db0cd54441f6b8366d10121d03d1

olya-shyian (changes requested):
Good job!
Unfortunately, the modifier does not work correctly on days...if you write Sunday, we will get this result

image

also note that the first column is Monday so if you write a modifier 'mon' then the calendar should start from the first column

test by specifying different days

DanilWeda (changes requested):
Please take a look comments, which writes other mentors. Also, I had hover which goes to the viewport border.

image

YegorKorenkov (approved):
You forgot align calendar to center, but everything is okey. Approve!

PR #2956, opened 2023-12-04
calendar created

DenisFrankovskyi:
https://denisfrankovskyi.github.io/layout_calendar/

Not sure if my @for and @each are ok, can you pelase check and advise if somethin should be corrected?


maxim2310 (changes requested):
you are going in correct way, but need to add modificators for calendar for it changes when i change this class (for example class="calendar calendar--start-day--sat calendar--month-length--29")

PR #2959, opened 2023-12-07
add calendar

alexsinchuk:
- [DEMO LINK](https://alexsinchuk.github.io/layout_calendar/)
- [TEST REPORT LINK](https://alexsinchuk.github.io/layout_calendar/report/html_report/)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar">
+      <div class="calendar__block calendar__block--1 calendar__block--1--sun"></div>
```
```suggestion
      <div class="calendar__block"></div>
```
remove modificators for block

use `:nth-child()` instead of `--1`

also, modificator like `--sun` should be added to `calendar`, not to the `calendar__block` element

the same with `calendar__block--last` - redundant:
image

use CSS combinators and `:nth-child()` to solve it

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+.calendar {
+  margin: 100px auto;
+  padding: $padding;
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap;
+  width: ($block-size * $column-count) + ($gap * $column-count) + $padding;
+
+  &__block {
+    box-sizing: border-box;
```
TarasHoliuk: it's better to set for each element on the page (use `*` selector)
note that it's only property allowed to set there

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
+$block-size: 100px;
+$gap: 1px;
+$column-count: 7;
+$padding: 10px;
+
 body {
   margin: 0;
 }
+
+.calendar {
```
TarasHoliuk: move this BEM-block to a separate file

TarasHoliuk (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2968, opened 2023-12-09
layout_calendar

DmitryVolchokFE:
- [DEMO LINK](https://DmitryVolchokFE.github.io/layout_calendar/)
- [TEST REPORT LINK](https://DmitryVolchokFE.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,8 +7,8 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="calendar calendar--30 ">
-      <div class="calendar__day calendar__day--thu"></div>
+    <div class="calendar calendar--30 calendar--tue">
+      <div class="calendar__day"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,6 +7,38 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--30 ">
+      <div class="calendar__day calendar__day--thu"></div>
```
Esceype: It works well but I think it will be better to set modifier for whole calendar for block, cause it change the look not only for the current one.

Esceype (changes requested):
Great job! Everything works good, but read the following comment, I think it will be better way.

anastasiiavorobiova (approved):
Great work!

PR #2985, opened 2023-12-15
add task solution calendar

Safina-coder:
- [DEMO LINK](https://Safina-coder.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Safina-coder.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,62 +1,51 @@
-$days: tue 1, wed 2, thu 3, fri 4, sat 5, sun 6;
-$cell-size: 100px;
-$calendar-padding: 10px;
-$calendar-gap: 1px;
+@import 'variables';
 
-* {
-  box-sizing: border-box;
+body {
   margin: 0;
 }
 
 .calendar {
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+    @for $newDay from 1 through $day {
+      .calendar__day:nth-child(#{$newDay}) {
+        visibility: visible;
+        &::before {
+          content: $newDay + '';
+        }
+      }
+    }
+  }
+}
```
Esceype: This modifier breaks if we put a number less than 28 or more than 31, and it should work as follows, if we put more than 31 or less than 28, the modifier should simply not work and all dates should be visible.

Esceype (changes requested):

andriy-shymkiv (changes requested):
demo link isn't working properly. please fix it and request a review again

anastasiiavorobiova (approved):
Great work!

PR #2988, opened 2023-12-18
add solution

StageUa:
- [DEMO LINK](https://stageua.github.io/layout_calendar/)
- [TEST REPORT LINK](https://stageua.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,63 @@
+@import "/styles/variables";
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap-size;
+  padding: 10px;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+      &::nth-child(n + #{$i + 1}) {
+        display: none;
+      }
+    }
+
+    &:hover {
+      background-color: pink;
+      cursor: pointer;
+      transform: translateY(-20px);
+      transition: all 0.5s ease;
```
ghost: The transition property should be declared on the original element, not on the hovered state. This way, the transition will also apply when the mouse leaves the element.

src/styles/main.scss
```diff
@@ -1,3 +1,63 @@
...
+    }
+  }
+}
+
+$days: mon 0, tue 1 wed 2, thu 3 fri 4, sat 5, sun 6;
```
ghost: You can put this variable in the "variables" file.

ghost (changes requested):
Good job! See the comments and set the days in the calendar as it stated in the task!

ghost (approved):
Great job!

PR #3073, opened 2024-01-17
Calendar

Vovan4ik03:
- [DEMO LINK](https://vovan4ik1.github.io/layout_calendar/)
-  [ ] Changing 'month-lengh' and 'start-day' modifier in the code element
- [ ] Each day has no modifiers, only class (eg. calendar__day)
- [ ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ ] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,45 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
image
image
looks like different names :)

PR #3090, opened 2024-01-22
Develop

Vovan4ik03:
- [DEMO LINK](https://vovan4ik1.github.io/layout_calendar/)
-

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,13 +8,7 @@
   </head>
   <body>
 
-    <div
-    class="
-    calendar
-    calendar--start-day-wed
-    calendar--mouth-lenght-29
-    "
-    >
+    <div class="calendar calendar--start-day-wed calendar--mouth-lenght-29">
 
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -7,6 +7,45 @@
...
   <body>
-    <h1>Calendar</h1>
+
+    <div
+    class="
+    calendar
+    calendar--start-day-wed
+    calendar--mouth-lenght-29
+    "
+    >
```
```suggestion
    <div class="calendar calendar--start-day-wed calendar--mouth-length-29">
```
the length should be less than 80 characters in one row

Also, typo not ~lenght~ but **length**

src/styles/main.scss
```diff
@@ -1,3 +1,59 @@
...
+  font-family: $calendar-font;
+  font-size: 30px;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100vh;
+}
+
+.calendar {
```
TarasHoliuk: let's move it to a separate file (according to BEM rules - it's a separate block)

TarasHoliuk (changes requested):
Well done - 2 small comments

PR #3127, opened 2024-02-01
Add solution

Vasya564:
- [DEMO LINK](https://Vasya564.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Vasya564.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,61 @@
+@import './variables';
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
   margin: 0;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $day-cell-gap;
+  width: $row-width;
+
+  &__day {
+    position: relative;
+    display: none;
+    box-sizing: border-box;
+    min-width: $day-cell-size;
+    height: $day-cell-size;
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+  &__day {
+    position: relative;
+    display: none;
+    box-sizing: border-box;
+    min-width: $day-cell-size;
+    height: $day-cell-size;
+    border: 1px solid black;
+    background-color: #eee;
+
+    transition: all 0.5s;
```
BudnikOleksii: Don't set transition to all properties, specify to which properties it should apple, in your case `transform` and `background-color`

src/styles/main.scss
```diff
@@ -1,3 +1,62 @@
...
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: '#{$i}';
+        position: absolute;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
+        font-size: 30px;
```
BudnikOleksii: - All of them have the same styles, so apply them only once instead of doing it in the loop.
- And it's not a good practice to center with absolute position, you can just center with flex for the parent container

BudnikOleksii (changes requested):
Looks good 👍 
But could be even better, check the comments I left a few suggestions.

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #3129, opened 2024-02-01
calendar done

anastasia1383:
- [DEMO LINK](https://anastasia1383.github.io/layout_calendar/)
- [TEST REPORT LINK](https://anastasia1383.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,7 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body class="body">
+  <body>
     <div class="calendar calendar--start-day-sun calendar--month-length-31">
       <div class="calendar__day"></div>
       <div class="calendar__day"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="body">
+    <div class="calendar calendar--start-day-sun calendar--month-length-31">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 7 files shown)


src/styles/calendar.scss
```diff
@@ -0,0 +1,46 @@
...
+    height: $day-size;
+    box-sizing: border-box;
+    background-color: $main-bg-color;
+    border: 1px solid $main-black-color;
+    transition: all 0.5s ease;
+
+    @for $days from 1 through 31 {
+      &:nth-child(#{$days})::before {
+        content: "#{$days}";
+        margin: auto;
```
BudnikOleksii: - Specify to which properties should transition apply, and actually now it doesn't work
- `margin: auto;` is the same for all children of the day, so you can specify only once instead of doing it in a loop. And, also it doesn't center now
https://www.loom.com/share/05415df45a3143c2850acf25ad97f036?sid=8eb7d7dc-73a9-412d-bde5-f7ab37a94830

BudnikOleksii (changes requested):
Good job 👍 
But a few moments should be fixed, I left a comment with a video recording.

anastasia1383:
Thanks for the review!

StasSokolov1 (changes requested):
Almost done but the issue with transition still exists check, please

anastasia1383:
Thank you, I fixed the demo link.

src/index.html
```diff
@@ -6,7 +6,39 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="body">
```
dffUqp:
```suggestion
  <body>
```

src/styles/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+        display: none;
+      }
+    }
+  }
+
+  &__day {
+    position: relative;
+    width: $day-size;
+    height: $day-size;
+    box-sizing: border-box;
```
dffUqp: it s better to apply this property to all elements (use * selector)

When all elements have the same box-sizing value, their behavior becomes more predictable

src/styles/calendar.scss
```diff
@@ -0,0 +1,49 @@
...
+    background-color: $main-bg-color;
+    border: 1px solid $main-black-color;
+    transition: transform, background-color 0.5s ease;
```
dffUqp, line 30: I think this transition doesn't work for `::before`

dffUqp (changes requested):
Almost done🙂

src/styles/calendar.scss
anastasia1383: how should it work for the before element?

BudnikOleksii: > how should it work for the before element?

`transition: transform  0.5s ease, background-color 0.5s ease;`
https://dev.to/clairecodes/css-transitions-for-multiple-properties-on-one-element-9o

src/styles/calendar.scss
```diff
@@ -0,0 +1,48 @@
...
+      cursor: pointer;
+    }
+
+    @for $days from 1 through 31 {
+      &:nth-child(#{$days})::before {
+        content: "#{$days}";
+        position: absolute;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
```
BudnikOleksii, line 44: - These properties are the same, so you don't need to specify them inside loop.
- Also, you can center with flex for the parent container instead of position absolute.

BudnikOleksii (approved):
Approved and left a few suggestions for the future

PR #3132, opened 2024-02-01
solution

ChrisHryts:
- [DEMO LINK](https://chrishryts.github.io/layout_calendar/)
- [TEST REPORT LINK](https://chrishryts.github.io/layout_calendar/report/html_report/)

- [x] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,79 @@
+$month-length: 31;
+$calendar-padding: 10px;
+$box-width: 100px;
+$box-height: 100px;
+$box-gap: 1px;
+$font-size: 30px;
+$days: (0: 'mon', 1: 'tue', 2: 'wed', 3: 'thu', 4: 'fri', 5: 'sat', 6: 'sun');
+
+html {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
   margin: 0;
+  
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  gap: $box-gap;
+  flex-wrap: wrap;
+
+  box-sizing: border-box;
+  padding: $calendar-padding;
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
   margin: 0;
+  
+  height: 100vh;
+}
+
+.calendar {
```
BudnikOleksii: Don't put all in one files, create a separate file for each BEM block

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+
+  &__day {
+    box-sizing: border-box;
+    height: $box-height;
+    width: $box-width;
+    
+    background-color: #eee;
+    border: 1px solid #000;
+
+    transition: all 0.5s;
```
BudnikOleksii: Apply transition only for properties that actually change instead of all

src/styles/main.scss
```diff
@@ -1,3 +1,79 @@
...
+    @for $i from 1 through $month-length {  
+      &:nth-child(#{$i})::before {
+        content: '#{$i}';
+
+        display: flex;
+        justify-content: center;
+        line-height: $box-height;
+
+        font-size: $font-size;
+        color: #000;
```
BudnikOleksii: These properties are the same for all of them, so specify just once instead of doing it in the for loop

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution. Also, deploy it so the next reviewer can check the UI:
image

src/styles/main.scss
ChrisHryts: fixed

src/styles/main.scss
ChrisHryts: fixed

src/styles/main.scss
ChrisHryts: fixed

StasSokolov1 (approved):
Looks good to me🔥

PR #3158, opened 2024-02-02
TaskComplete

TideTwo:
(My code fails the tests, but the readme says there is no tests, which i assume is due to the gap requirement)

```- This task does not have tests so you need to check if it works as expected manually.```

- [DEMO LINK](https://tidetwo.github.io/layout_calendar/)
- [TEST REPORT LINK](https://tidetwo.github.io/layout_calendar/report/html_report/)

- [x] Changing 'month-length' and 'start-day' modifier in the code element

### CODE REVIEW

File: src/styles/main.scss [hunk 1]
```diff
@@ -1,3 +1,81 @@
...
+);
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap-size;
+
+  // added border to see padding
+
+  border: 1px solid $black-color;
```
asakevych: because of the border the tests are not passed

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
...
+$black-color: #000;
+$days: (
+  "mon": 0,
+  "tue": 1,
+  "wed": 2,
+  "thu": 3,
+  "fri": 4,
+  "sat": 5,
+  "sun": 6,
+);
```
asakevych: create separate file for variables

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
...
+
+.calendar {
```
asakevych: block calendar should be in separate file

src/styles/main.scss
```diff
@@ -1,3 +1,81 @@
 body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
   margin: 0;
+  height: 100vh;
+}
```
asakevych: you can use class to add some specific styles to body and some resets as `margin: 0` should be in reset.css

asakevych (changes requested):
Great job!! :rocket: :fire: just change a little bit a structure of your files

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,6 @@
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
```
asakevych: it's better to use `.page` for styling here if you have page block

src/styles/reset/reset.scss
```diff
@@ -0,0 +1,3 @@
+.page {
```
asakevych: here you can use tag body, in files like reset.scss or normalize.scss we usually give styles for tags, not for classes

asakevych (changes requested):
It looks good, but still there are small things to fix
I think you didn't do deploy, I cannot see any changes on your page and tests are still not passed
please remember about deploying changes and make sure tests are passed

TideTwo:
Apologies, I had assumed the site gets redeployed on pushing another commit, like tests are run automatically. Also,Im not sure how I swapped around the body and .page selectors, fixed now.

src/styles/reset/reset.scss
```diff
@@ -0,0 +1,3 @@
+body {
+  margin: 0;
+}
```
asakevych: add 
```suggestion
* {
  box-sizing: border-box
}
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,55 @@
+@use "../variables/variables" as *;
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap-size;
+  padding: $padding-size;
+  width: $day-size * 7 + $padding-size;
```
asakevych: 1. if you want to calculate paddings into width you should add `box-sizing: border-box`, you can add it for all elements using `*` 
2. you have 2 paddings, and you have 1px gap between every day
so you need to calculate `($day-size * 7) + ($padding-size * 2) + ($gap * 6)`

asakevych (changes requested):
please make sure tests are passed, I still see your tests are not passed, try to change width settings for calendar, see my comments :arrow_down:

src/styles/blocks/calendar.scss
TideTwo: As stated in the original pr comment, the task readme specifically states that the tests are not relevant. Because of that, I ignored the tests failing. I can fix up the project to pass the tests, but could you update the readme to show we need to take the test into consideration? Or at least confirm that they need to be passed? I have followed all the requests stated by the task, so I guess I'll skip this task till I get a response.

nataliaklonowska (approved):

PR #3162, opened 2024-02-05
task done

Yan-Chan-py:
https://yan-chan-py.github.io/layout_product-cards/
https://<your_account>.github.io/layout_product-cards/report/html_report/

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -1,0 +1,53 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100vh;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  padding: 10px;
+  gap: $gap;
+  width: $day-rows * $box-size + ($day-rows - 1) * $gap;
+
+  @each $day, $day-number in $days {
+    &--start-day-#{$day} &__day:first-child {
+      margin-left: $day-number * ($box-size + $gap);
+    }
+  }
+
+  @for $i from 28 through 31 {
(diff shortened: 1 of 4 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
+body {
+  margin: 0;
+  font-family: Arial, sans-serif;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100vh;
+}
```
dffUqp, line 8: Let's move this styles to `main.scss` file

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: $box-size;
+    height: $box-size;
+    box-sizing: border-box;
```
dffUqp, line 35: You shouldn't use `*` selector for `margin` and `padding`, because it impacts the performance. 

But what about `box-sizing`? 🤔

In case you have 30 different blocks on the page, will you write `box-sizing` to all these blocks? In a big project if somebody forgot to add `box-sizing` it can be hard to debug in the future.

So it's good to write `box-sizing` in `*` selector. Let's move it there.

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    font-size: 30px;
+    background-color: $box-color;
+    border: solid 1px $box-border-color;
+    transition: all 0.5s;
+
+    &:hover {
+      cursor: pointer;
```
dffUqp, line 42: Let's move `cursor: pointer;` style to the `&__day` selector

dffUqp (changes requested):
LGTM 🔥 but needs a few minor fixes 🙂 

Please, check your Demo link and tests

PR #3225, opened 2024-02-23
Implement calendar

olafchuszno:
- [DEMO LINK](https://olafchuszno.github.io/layout_calendar/)
- [TEST REPORT LINK](https://olafchuszno.github.io/layout_calendar/report/html_report/) 

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,44 @@
     <title>Calendar</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <body class="page-content">
+    <div class="
+      calendar
+      page-content__calendar
+      calendar--start-day-sun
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -7,6 +7,42 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="
```
yevhenii-pyl, line 10: No way a class name can be written like this. It's a valid approach when you use classNames package (example below), but not here. 

```
classNames({
		btn: true,
		'btn-pressed': isPressed,
		'btn-over': !isPressed && isHovered,
	});
```

yevhenii-pyl (approved):
Solid 🚀 

Just one note on a class name at the html, but it's just 'for future', see no reason to block PR.

PR #3275, opened 2024-03-08
add solution Artem Zakharchuk

ArtemZakhar:
Smth strange happening with my gap distance. I don't know how to fix.
Please advice if there can be solution.

- [DEMO LINK](https://ArtemZakhar.github.io/layout_calendar/)
- [TEST REPORT LINK](https://ArtemZakhar.github.io/layout_calendar/report/html_report/)

- [X] Changing 'month-lengh' and 'start-day' modifier in the code element
- [X] Each day has no modifiers, only class (eg. calendar\_\_day)
- [X] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [X] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,13 +6,119 @@
       name="viewport"
       content="width=device-width, initial-scale=1.0"
     />
+
     <title>Calendar</title>
+
     <link
       rel="stylesheet"
-      href="styles/index.scss"
+      href="styles/main.scss"
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
+      <div class="calendar calendar--month-length-31 calendar--start-day-sun">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+      </div>
+    </div>
+    <div class="wrapper">
+      <div class="calendar calendar--month-length-29 calendar--start-day-sun">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+      </div>
+    </div>
+    <div class="wrapper">
+      <div class="calendar calendar--month-length-31 calendar--start-day-wed">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -6,13 +6,119 @@
...
+
     <link
       rel="stylesheet"
-      href="styles/index.scss"
+      href="styles/main.scss"
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="wrapper">
```
mrkovaliv, line 18: Actually you don't need `<div class="wrapper">` here you can use calendar like your wrapper block

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap;
+  box-sizing: border-box;
```
mrkovaliv: Better move this logic to reset.scss and import here

```
html {
  box-sizing: border-box;
}

*,
*::after,
*::before {
  box-sizing: inherit;
}

body {
  margin: 0;
}
```

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
+@import './utils/variables';
+
+$padding: 10px;
```
mrkovaliv: move all these variables to variables.scss but rename them to make more specify

example:
$padding -> $calendar-padding

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+$gap: 1px;
+$total-gap: $gap * 8;
+$square-side: 100px;
+$line-width: ($square-side * 7) + ($padding * 2) + $total-gap;
```
mrkovaliv: just `calendar-width`

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+  ),
+);
+
+body {
+  margin: 0;
+}
+
+.wrapper {
+  width: 100vw;
+  height: 100vh;
```
mrkovaliv: better use min-height: 100vh

`Using min-height: 100vh; allows for flexible content display, adapting to varying viewport sizes and accommodating dynamic content. It prevents overflow issues and enhances accessibility by ensuring all users can access content without encountering layout problems. Additionally, it future-proofs your design, allowing for seamless adjustments as your website evolves without the need for manual height adjustments.`

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+    ($square-side + $gap) * 6,
+  ),
+);
+
+body {
+  margin: 0;
+}
+
+.wrapper {
+  width: 100vw;
```
mrkovaliv: you can remove this

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+    transition: transform 0.5s;
+    &:hover {
+      cursor: pointer;
+      background-color: $pink;
+      transform: translateY(-20px);
+    }
+  }
+}
+
+@for $i from 28 through 31 {
```
mrkovaliv, line 42: your approach is not bad but there is better:

variables:


```
$calendar-cell-size: 100px;
$calendar-gap-size: 1px;
$calendar-padding: 10px;
$calendar-days: (
  mon: 0,
  tue: 1,
  wed: 2,
  thu: 3,
  fri: 4,
  sat: 5,
  sun: 6,
);
```
code:

```
  &__day {
    @for $i from 1 through 31 {
      &:nth-child(#{$i})::before {
        content: '#{$i}';
      }
    }
  }
```
```

  @each $calendar-day, $calendar-index in $calendar-days {
    &--start-day-#{$calendar-day} :first-child {
      margin-left: calc(
        #{$calendar-index} * $calendar-cell-size + #{$calendar-index} * $calendar-gap-size
      );
    }
  }

  @for $length from 28 through 31 {
    &--month-length-#{$length} :nth-child(n + #{$length + 1}) {
      display: none;
    }
  }

```

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+  .calendar--month-length-#{$i} {
+    @for $j from 1 through $i {
+      & :nth-child(#{$j}) {
+        display: block;
+        position: relative;
+        &::before {
+          content: '#{$j}';
+          font-size: 30px;
+          font-family: Arial, sans-serif;
```
mrkovaliv: move this style for calendar not just for day `font-family: Arial, sans-serif;`

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+
+@for $i from 28 through 31 {
+  .calendar--month-length-#{$i} {
+    @for $j from 1 through $i {
+      & :nth-child(#{$j}) {
+        display: block;
+        position: relative;
+        &::before {
+          content: '#{$j}';
+          font-size: 30px;
```
mrkovaliv: add this style for `day` not just for `&::before`

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+    position: relative;
+    display: none;
+    box-sizing: border-box;
+    padding: 0;
+    margin: 0;
+    width: $square-side;
+    height: $square-side;
+    background-color: $grey;
+    border: 1px solid #000;
+    transition: transform 0.5s;
```
mrkovaliv, line 31: add empty line to separate styles

src/styles/main.scss
```diff
@@ -0,0 +1,97 @@
...
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap;
+  box-sizing: border-box;
+  padding: $padding;
+  width: $line-width;
+
+  transform: translateX(1px);
```
mrkovaliv, line 19: why you added this style?, you don't need it :)

mrkovaliv (changes requested):
Nice work 👍  but I added some improvements

mrkovaliv:
I think after you correct everything I described, the tests won't be failed :)

src/styles/main.scss
ArtemZakhar: I was sure that this task would have better results, truly)
But still, thank you for closely watching!)

I won't pass any tests without this transition(

src/styles/main.scss
ArtemZakhar: If you don't mind I will leave my variant) 
Your approach is easier to read) 

But it was a challenge for me to use loop inside loop correctly here) 
It is hard to refuse my attempt)

src/styles/main.scss
ArtemZakhar: @mrkovaliv suddenly I realized that I missed asking you about my test failure)
so one thing is that I had to transform the whole calendar to pass 3 tests.
and another is about the flex gap, I suppose, please see the attached photo. 

It looks like every next section increased the difference from the reference because of the flex gap. 

Should I worried so about tests?

photo_2024-03-08_11-30-33

src/index.html
ArtemZakhar: @mrkovaliv I don't know how to make it real. 

we need the calendar to be positioned in the center.
I can't use absolute position because there are several calendars. 
I can make left margin with help of "margin: 0 auto". 
But what should I do with top margin?

DanilWeda (approved):
Good. 
All functionality works fine. 
I approved this request. 
Thanks.

src/styles/main.scss
mrkovaliv: no, everything works fine 👍

mrkovaliv (approved):

PR #3305, opened 2024-03-15
calendar solution

AYupko:
- [DEMO LINK](https://AYupko.github.io/layout_calendar/)
- [TEST REPORT LINK](https://AYupko.github.io/layout_calendar/report/html_report/)


src/index.html
```diff
@@ -13,6 +13,40 @@
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="container">
+      <div class="calendar calendar--start-day-sun calendar--month-length-31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
```

DanilWeda (changes requested):
Great! 
All good, please check small comment and fix test suite. 

Many thanks)

src/index.html
AYupko: Any hint how I can deal with tests? I put "gap: 1px" in flex container, but it doesn't work correct.

undeletable: @AYupko Let's take a look at the failing tests.
Two tests are failing in case when month starts from Wednesday or has 29 days:
calendar-failure-5rows
In both cases there are 5 rows in the calendar, unlike the example you're using in your code (31 day starting from Sunday => 6 rows). 6 rows completely fill the height of calendar, while with 5 rows there;s some free space, distributed between the rows. See how it looks like in 'TEST' column.
In this case I suggest to remove height style from `.calendar`, so that these rows would not be distributed vertically.
When I've done that, I've noticed that gap between rows is not applied. I've noticed that this was caused by the way you position the `.container`. Looks like the CSS transformation breaks calculation of gap between rows.
I suggest you to remove the existing styles for `.container`, and instead add the following ones: full height (100vh), width equal to calendar width (I tried using full width, but somehow tests fail), and also define it as flex container with contents centered vertically and horizontally.
The only thing left would be to center `.container` relatively to its parent. If you don't want to style its parent as flex container, you can just add `margin: 0 auto` style to `.container`. Effectively, this centers `.container` horizontally, while vertical centering is not needed as it occupies full height (more: https://learnlayout.com/margin-auto).
With these changes, tests are passing. Goog luck!

AYupko: @undeletable Thank you very much for such a detailed explanation

DanilWeda (approved):
Great.
Approved.
Many thanks)

PR #3306, opened 2024-03-15
Develop

yanamomot:
[DEMO LINK](https://yanamomot.github.io/layout_calendar/)
[TEST REPORT LINK](https://yanamomot.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,39 +15,39 @@
 
   <body>
     <div class="main-box">
-      <div class="calendar calendar--start-day-sun calendar--month-length-31">
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
-        <div class="calendar__day"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -12,7 +12,42 @@
...
     />
   </head>
+
   <body>
-    <h1>Calendar</h1>
+    <div class="main-box">
+      <div class="calendar calendar--start-day-sun calendar--month-length-31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
```
In this case best variant uses list)

src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+$gap: 1px;
+$weekdayIter: (
+  mon 0,
+  tue 101px,
+  wed 202px,
+  thu 303px,
+  fri 404px,
+  sat 505px,
+  sun 606px
+);
```
dothearts: Move to separate variables file

DanilWeda (changes requested):
Good job! 
Check small comments) 
Thanks!

yanamomot:
Hi. I replace all div to ul li and add new scss file for values. But Kyrylo said what we can use div, because it makes no difference :)

src/index.html
yanamomot: Hi. I replace all div to ul li and add new scss file for values. But Kyrylo said what we can use div, because it makes no difference :)

DanilWeda (approved):
Great!
Approved.
Thanks)

PR #3308, opened 2024-03-15
Complete with calendar but..

sevenlays:
- [DEMO LINK](https://sevenlays.github.io/layout_calendar/)
- [TEST REPORT LINK](https://sevenlays.github.io/layout_calendar/report/html_report/)

[x] Changing 'month-lengh' and 'start-day' modifier in the code element
- [x] All `Typical Mistakes` from `BEM` lesson theory are checked. ?
- [x] Code follows all the [Code Style Rules ❗️] ?

If you would be so kind as to approve my code, could you please point out any errors I may have made? Please be extremely meticulous. Also if you reject)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,3 +1,4 @@
+<!-- #region -->
 <!doctype html>
 <html lang="en">
   <head>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -12,7 +13,43 @@
       href="styles/index.scss"
     />
   </head>
-  <body>
-    <h1>Calendar</h1>
+  <!-- #endregion -->
```
dothearts: Do you need this commented line?

src/index.html
```diff
@@ -12,7 +13,43 @@
...
+  <body class="page">
+    <!-- <h1 class="hidden">Calendar</h1> -->
```
dothearts: Do you need this commented line?

src/index.html
```diff
@@ -12,7 +13,43 @@
...
+    <div class="container">
+      <div class="calendar calendar--start-day-sun calendar--month-length-31">
+        <div class="calendar__day"></div>
+        <div class="calendar__day"></div>
```

DanilWeda (changes requested):
All good! 
Check small comments. 
Thanks)

src/index.html
sevenlays: and  For convenience only:)

mrkovaliv (approved):
Nice work 👍

PR #3309, opened 2024-03-15
Calendar

TetianaVeremchuk:
- [DEMO LINK](https://tetianaveremchuk.github.io/layout_calendar/)
- [TEST REPORT LINK](https://tetianaveremchuk.github.io/layout_calendar/report/html_report/)

-[x] Changing 'month-lengh' and 'start-day' modifier in the code element reflects in changing calendar layout
-[x] Each day has no modifiers, only class (eg. calendar__day)
-[x] All Typical Mistakes from BEM lesson theory are checked.
-[x] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,38 @@
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sun calendar--month-length-31">
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
+      <div class="calendar__day"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -13,6 +13,38 @@
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sun calendar--month-length-31">
+      <div class="calendar__day"></div>
```
dothearts, line 17: Try to use ul and li in this case.

DanilWeda (approved):
Great.
Approved with small comment) 
Thanks!

PR #3316, opened 2024-03-17
add solution

stepanyukvova200:
- [DEMO LINK](https://stepanyukvova200.github.io/layout_calendar/)
- [TEST REPORT LINK](https://stepanyukvova200.github.io/layout_calendar/report/html_report/)

Also, why can't I use 'index (list, value)'?


alexander-ignatow (changes requested):
> Also, why can't I use 'index (list, value)'?
Did not get your question

Also - can you please fix the tests (they are failing because of calendar block padding)

alexander-ignatow (changes requested):
Did not get your question
> Also, why can't I use 'index (list, value)'?

Also - can you please fix the tests (they are failing because of calendar block padding)

The code itself looks good!

stepanyukvova200:
ok, all tests have passed )

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,53 @@
...
+    background-color: #eee;
+
+    border: 1px solid #000;
+
+    transition: transform 0.5s ease;
+
+    &:hover {
+      cursor: pointer;
+
+      // background-color: #ffbfcb; in the test background is grey =(
```
alexander-ignatow, line 22: In the future - it is a bad practice to keep the commented-out code. If you need it - try transforming it into some meaningful comment. If it is not needed - just remove it.

alexander-ignatow (approved):
Looks good 👍 

Just mind my comment for the future 👇

PR #3337, opened 2024-03-27
layout_calendar

BagietaCyganeta:
- [DEMO LINK](https://BagietaCyganeta.github.io/layout_calendar/)
- [TEST REPORT LINK](https://BagietaCyganeta.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,70 @@
+$calendar-width: 726px;
+
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  font-size: 30px;
+  font-family: Arial, sans-serif;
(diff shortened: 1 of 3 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+  width: $calendar-width;
+  padding: 10px;
+}
+
+$day-of-week: (
```
yevhenii-pyl, line 26: Interesting approach with indexes. :) 
Why have you used three digits for index, instead of just one? 
I mean, looks like just 0 - 6 would have worked just fine.

yevhenii-pyl (approved):
Solid!
I can suggest creating a separate file for scss variables, just to structure your code better, but that's not the reason to block your PR here. 
Keep pushing 🥇

PR #3365, opened 2024-04-10
add solution for calendar task

Oleksii-Bidiak:
- [DEMO LINK](https://Oleksii-Bidiak.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Oleksii-Bidiak.github.io/layout_calendar/report/html_report/)


src/index.html
```diff
@@ -13,6 +13,131 @@
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-sun calendar--month-length-31">
+      <a
+        href="#"
+        class="calendar__day"
+      ></a>
```
volodymyr-soltys97: Fix it everywhere
```suggestion
      <div class="calendar__day"></div>
```

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,57 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap;
+  padding: $paddingCalendarBox;
+  width: $calendarSize;
+  box-sizing: border-box;
```
volodymyr-soltys97: Move box-sizing to the `*` selector and remove it in other places

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
This part of the task is incomplete
image
image

src/styles/blocks/calendar.scss
Oleksii-Bidiak: But the code of the layout_task-guideline style says that using the * selector is not advisable. Can I deviate from this rule?

src/styles/global/general.scss
```diff
@@ -0,0 +1,8 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
DarkMistyRoom, line 4: body doesn't have default padding

src/styles/global/variables.scss
```diff
@@ -0,0 +1,20 @@
...
+$gap: 1px;
+$calendarSize: $paddingCalendarBox * 2 + $gap * 6 + $widthCell * 7;
+$start-day: (
+  'mon': 0,
+  'tue': $widthCell + $gap,
+  'wed': $widthCell * 2 + $gap * 2,
+  'thu': $widthCell * 3 + $gap * 3,
+  'fri': $widthCell * 4 + $gap * 4,
+  'sat': $widthCell * 5 + $gap * 5,
+  'sun': $widthCell * 6 + $gap * 6,
```
DarkMistyRoom: looks good, but we can use a formula to avoid duplications

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,56 @@
...
+    border: 1px solid black;
+
+    height: $heightCell;
+    width: $widthCell;
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    transition: all 0.5s;
```
DarkMistyRoom: list transition properties instead of using `all`, it is better for performance and move 0.5s to a variable

src/styles/blocks/calendar.scss
DarkMistyRoom: It is ok to use * for box-sizing

DarkMistyRoom (changes requested):
Good job! But cursor should become pointer over the calendar cells

etojeDenys (approved):
lgtm

PR #3372, opened 2024-04-12
commit

SerhiiNalyvaiko:
- [DEMO LINK](https://SerhiiNalyvaiko.github.io/layout_calendar/)
- [TEST REPORT LINK](https://SerhiiNalyvaiko.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/var.scss
+++ b/src/styles/var.scss
@@ -1,0 +1,16 @@
+$weekdays: (
+  'mon': 0,
+  'tue': 1,
+  'wed': 2,
+  'thu': 3,
+  'fri': 4,
+  'sat': 5,
+  'sun': 6,
+);
+$gap: 1px;
+$squareDays: 100px;
+$heightDay: 100px;
+$widthDay: 100px;
+$daysInRow: 7;
+$mounthLength: 31;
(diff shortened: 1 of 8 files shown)


src/styles/var.scss
```diff
@@ -0,0 +1,16 @@
...
+  'wed': 2,
+  'thu': 3,
+  'fri': 4,
+  'sat': 5,
+  'sun': 6,
+);
+$gap: 1px;
+$squareDays: 100px;
+$heightDay: 100px;
+$widthDay: 100px;
```
volodymyr-soltys97: No need to duplicate variables with same values
```suggestion
$daySize: 100px;
```

volodymyr-soltys97 (changes requested):
Hi, the modifier for the beginning of the day of the week does not work correctly, we should always have 7 columns
image

vadiimvooo (changes requested):
Seems like your deploy is broken, please redeploy your solution. Feel free to ask for help in fe_chat if you need it.

etojeDenys (approved):
GJ

PR #3374, opened 2024-04-15
solution

Teslaanazuma:
Display a calendar in the middle of the screen (both horizontally and vertically). Use SCSS and follow BEM. Don't use JS.

Write styles in src/styles/main.scss instead of src/style.css.
Create a markup for the calendar block with 31 days inside
DON'T add numbers in HTML (you will do it using CSS)
Each day is a grey (#eee) 100px square (including 1px black border)
Add a number (Arial 30px) in the center of each day using ::before and [@for](https://sass-lang.com/documentation/at-rules/control/for)
Use flex with 1px gap and limit its width to exactly 7 columns + 10px paddings
Don't use hardcoded px values if they are used several times
Use properly named variables to make all the calculations more clear.
Implement start-day modifier for the calendar with mon, tue, wed, thu, fri, sat and sun values
Use [@each](https://sass-lang.com/documentation/at-rules/control/each) to create all the modifiers
The month should start at the correct column (Monday is the 1st, Friday is the 5th)
You can just add correct margin-left for the first day
Set calendar to start from Sunday by default
Add a modifier month-length for the calendar with values 28, 29, 30 and 31 (use @for)
It sets the last day to show (use [nth-child](https://css-tricks.com/how-nth-child-works/))
Set 31 days by default
On hovering over a cell:

cursor should become pointer
The hovered cell has to become pink (use #FFBFCB)
Move the hovered cell up by 20px (use transform)
All changes should be animated with the duration of 0.5s

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ On hovering over a cell:
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_calendar/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_calendar/report/html_report/)
+- [DEMO LINK](https://Teslaanazuma.github.io/layout_calendar/)
+- [TEST REPORT LINK](https://Teslaanazuma.github.io/layout_calendar/report/html_report/)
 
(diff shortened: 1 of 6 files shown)


anastasiiavorobiova (changes requested):
Great work! Make sure to deploy your solution and provide a demo link in the PR description

anastasiiavorobiova (approved):
Awesome work!

PR #3424, opened 2024-05-03
add task solution

OSpyrydonov:
- [DEMO LINK](https://OSpyrydonov.github.io/layout_calendar/)
- [TEST REPORT LINK](https://OSpyrydonov.github.io/layout_calendar/report/html_report/)


src/styles/null.scss
dinaminahi, line 1: Not the best names for file, better would be for example variables.scss and main.scss
When you name it null, it difficult to understand what you have in it and we don't usually use shortcuts for names 
image

dinaminahi (approved):
Great job ⭐

PR #3531, opened 2024-06-20
add task solution

tuhusova:
- [DEMO LINK](https://tuhusova.github.io/layout_calendar/)
- [TEST REPORT LINK](https://tuhusova.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,76 @@
+$grey: #eee;
+$black: #000;
+$pink: #ffbfcb;
+$day-size: 100px;
+$border-size: 1px;
+$font-size: 30px;
+$gap-size: 1px;
+$padding-size: 10px;
+$animation-duration: 0.5s;
+
+// Основные стили для календаря
+body,
+html {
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,76 @@
...
+body,
+html {
+  height: 100%;
+  margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.calendar {
```
TarasHoliuk, line 21: According to BEM rules, each Block should be in a separate file
Same for variables

Also, no need to leave comments if the code is self-descriptive enough 🙂 But even if you left comments they should be in English

Generally, everything else looks good - approved ✔️

TarasHoliuk (approved):
P.S. Next time make sure deployed page work as expected before review

src/styles/main.scss
tuhusova: So in that case here should be 3 blocks: index, calendar and variables, right?

PR #3532, opened 2024-06-20
add task solution

YevychAndr:
- [DEMO LINK](https://YevychAndr.github.io/layout_calendar/)
- [TEST REPORT LINK](https://YevychAndr.github.io/layout_calendar/report/html_report/)


src/styles/main.scss
```diff
@@ -0,0 +1,67 @@
...
+    &--month-length-#{$length} :nth-child(n + #{$length + 1}) {
+      display: none;
+    }
+  }
+
+  @for $number from 1 through 31 {
+    &__day:nth-child(#{$number})::before {
+      content: '#{$number}';
+      font-family: Arial, sans-serif;
+      font-size: 30px;
```
TarasHoliuk: you will duplicate these properties 31 time, which is unnecessary
Font styles should be inherited from `&__day`. Or at least you can define it inside `&__day::before` but not inside the loop to avoid duplication

src/styles/main.scss
```diff
@@ -0,0 +1,67 @@
...
+    background-color: $day-color;
+    border: 1px solid black;
+    justify-content: center;
+    align-items: center;
+
+    &:hover {
+      cursor: pointer;
+      background-color: $day-color-hover;
+      transform: translateY(-20px);
+      transition:
```
TarasHoliuk: you should add a transition to the element not to element:hover to make it work in both ways

src/styles/main.scss
```diff
@@ -0,0 +1,67 @@
...
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+}
+
+body {
+  margin: 0;
+}
+
+.calendar {
```
TarasHoliuk: According to BEM rules, each BEM block should be in a separate file
Also, you can move variables to a separate file as well

TarasHoliuk (changes requested):
GJ 👍 
A few minor comments only

src/styles/variables.scss
```diff
@@ -0,0 +1,8 @@
+$padding: 10px;
+$gap: 1px;
+$day-color: #eee;
+$day-color-hover: #ffbfcb;
+$day-height: 100px;
+$day-width: 100px;
```
volodymyr-soltys97, line 6: Not need to duplicate variables with same values
```suggestion
$day-size: 100px;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #3543, opened 2024-06-21
all tasks are performed

NIKaragu:
- [DEMO LINK](https://NIKaragu.github.io/layout_calendar/)
- [TEST REPORT LINK](https://NIKaragu.github.io/layout_calendar/report/html_report/)
===============================================================================
===============================================================================

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -40,7 +40,7 @@
   @for $length from 28 through $days-quantity {
     &--month-length-#{$length} {
       .calendar__day:nth-child(n + #{$length + 1}) {
-        visibility: hidden;
+        display: none;
       }
     }
   }
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -29,18 +29,14 @@
   }
 
   @each $weekday, $day-num in $weekdays {
-    &--start-day-#{$weekday} {
-      .calendar__day:first-child {
-        margin-left: $day-num * ($day-width + $day-gap);
-      }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,42 @@
...
+
+    @for $day from 1 through $days-quantity {
+      &:nth-child(#{$day})::before {
+        content: '#{$day}';
+      }
+    }
+    &:hover {
+      background-color: #ffbfcb;
+      transform: translateY(-20px);
+      transition: 0.5s;
```
dothearts: You can move the transition on the element. In this case, you have smooth animation when you hover and unhover.  
Because, now, when I move the cursor outside the day element, the animation cuts.

Also, you need to write which property needs to be animated ( with transition). The default value is 'all' but it has side effects and bad performances.

DanilWeda (changes requested):
Good job. 
Check a small comment to improve your solution. 
Thanks)

src/styles/blocks/calendar.scss
NIKaragu: But accordingly to the readme.md I had to set 'all' ((((
image

I'm gonna add unhovering effect

NIKaragu: done

dothearts: @NIKaragu  _All changes should be animated..._ 
It's not about using `all` property. It is about - 0.5 duration for all elements which be animated.

DanilWeda (approved):
I approved, check my answer about all property. 
In future use please a specific property, not 'all'. 
Many thanks.

PR #3577, opened 2024-07-04
add task solution

RostyslavSharuiev:
- [DEMO LINK](https://RostyslavSharuiev.github.io/layout_calendar/)
- [TEST REPORT LINK](https://RostyslavSharuiev.github.io/layout_calendar/report/html_report/)

- [x] Each day has no modifiers, only class (eg. calendar\_\_day)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,5 +1,6 @@
 $cell-size: 100px;
 $gap-size: 1px;
+$transition-time: 0.5s;
 $days-in-week: 7;
 $days: (
   'mon': 0,
@@ -33,7 +34,9 @@ body {
     position: relative;
     background-color: #eee;
     border: 1px solid black;
-    transition: all 0.5s;
+    transition:
+      transform 0.5s,
+      background-color 0.5s;
 
     @for $i from 1 through 31 {
       &:nth-child(#{$i})::before {
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,68 @@
+$cell-size: 100px;
+$gap-size: 1px;
+$days-in-week: 7;
+$days: (
+  'mon': 0,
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,70 @@
...
+  padding: 20px;
+
+  &__day {
+    box-sizing: border-box;
+    height: $cell-size;
+    width: $cell-size;
+    position: relative;
+    background-color: #eee;
+    border: 1px solid black;
+    transition: all 0.5s;
```
BudnikOleksii: Specify to which properties should transition apply

src/styles/main.scss
```diff
@@ -0,0 +1,70 @@
...
+      &:nth-child(#{$i})::before {
+        content: '#{$i}';
+        font-family: Arial, sans-serif;
+        font-size: 30px;
+        line-height: 30px;
+        position: absolute;
+        text-align: center;
+        bottom: ($cell-size - 2 * $gap-size) / 2;
+        right: ($cell-size - 3 * $gap-size) / 2;
+        transform: translate(50%, 50%);
```
BudnikOleksii, line 51: - These properties are the same for each element, so move them outside the loop
- It's better to avoid position absolute, in a case when you need to center the element you can use display flex

BudnikOleksii (approved):
Good work 👍 
Let's improve your solution slightly

src/styles/main.scss
RostyslavSharuiev: As for text centering in ::before, unfortunately, my tests crashed when using flex, so I had to reinvent the wheel this way...
In general, if it is possible to pass the task by bypassing the tests, then absolute positioning and the bottom and right tasks can be removed (they only position the text)

PR #3581, opened 2024-07-04
add task solution

st74eet:
- [DEMO LINK](https://st74eet.github.io/layout_calendar/)
- [TEST REPORT LINK](https://st74eet.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,70 @@
-body {
-  margin: 0;
+$calendar-week-days-sum: 7;
+$calendar-days: (
+  'mon': 0,
+  'tue': 1,
+  'wed': 2,
+  'thu': 3,
+  'fri': 4,
+  'sat': 5,
+  'sun': 6,
+);
+$start-day: 'sun';
+$calendar-month-length: 31;
+$general-font-family: Arial, Helvetica, sans-serif;
+$general-font-size: 30px;
+$calendar-size: 100px;
+$calendar-padding: 10px;
+
+html {
+  font-family: $general-font-family;
+  font-size: $general-font-size;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: calc($calendar-size * 7 + $calendar-padding * 2);
+  padding: $calendar-padding;
+
+  &__day {
+    display: flex;
+    background-color: #eee;
+    border: 1px solid black;
+    height: 99px;
+    width: 99px;
+    margin: 1px;
+    text-align: center;
+    justify-content: center;
+    align-items: center;
+
+    &:hover {
+      cursor: pointer;
+      background-color: #ffbfcb;
+      transition: 0.5s;
+      transform: translate(0, -20px);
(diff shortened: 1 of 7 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+    }
+
+    @for $i from 1 through $calendar-month-length {
+      &:nth-child(#{$i})::before {
+        content: '' + $i;
+        display: block;
+      }
+    }
+
+    @for $i from 1 through $calendar-week-days-sum {
```
TarasHoliuk: you can use @each loop in this case and iterate directly on `$calendar-days`

src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+      &:first-child {
+        margin-left: ($calendar-size * 5 + $calendar-padding + 6px);
```
TarasHoliuk: so this margin is actually static - you don't use $i (or variable from each loop)

So changing modificator doesn't work:
image

src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+      }
+    }
+  }
+
+  @for $amount-days from 28 through 31 {
+    &--month-length-#{$amount-days}:nth-child(n + #{$amount-days + 1}) {
```
TarasHoliuk: it doesn't work

`&--month-length-#{$amount-days}:nth-child` means you trying to find the nth calendar instead of the nth day

add space before `:nth-child`. Or use a combined selector and explicitly define selector `&--month-length-#{$amount-days} &__day:nth-child(//rest of the code here)`

src/styles/index.scss
```diff
@@ -1,3 +1,70 @@
...
+html {
+  font-family: $general-font-family;
+  font-size: $general-font-size;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.calendar {
```
TarasHoliuk: According to BEM rules each BEM-block should be in a separate file

Also, you can move variables to a separate file

TarasHoliuk (changes requested):

anastasiiavorobiova (changes requested):
Screenshot 2024-07-05 at 15 10 34
Great work! Consider centering your calendar vertically. The container size should match the actual size of the calendar. Also, the first row is a bit shifted. Consider fixing it

src/styles/variables.scss
```diff
@@ -0,0 +1,17 @@
...
+$calendar-days: (
+  'mon': 1,
+  'tue': 2,
+  'wed': 3,
+  'thu': 4,
+  'fri': 5,
+  'sat': 6,
+  'sun': 7,
+);
+$start-day: 'sun';
```
etojeDenys, line 11: looks redundant

src/styles/variables.scss
```diff
@@ -0,0 +1,17 @@
+$calendar-week-days-sum: 7;
```
etojeDenys, line 1: same here

etojeDenys (approved):

etojeDenys (approved):

PR #3587, opened 2024-07-07
Develop

DmytroPAvlichenko:
- [DEMO LINK](https://DmytroPAvlichenko.github.io/layout_calendar/)
- [TEST REPORT LINK](https://DmytroPAvlichenko.github.io/layout_calendar/report/html_report/)


src/styles/index.scss
```diff
@@ -1,3 +1,65 @@
...
+$gap: 1px;
+
 body {
   margin: 0;
 }
+
+.calendar {
+  display: flex;
+  align-items: center;
+  position: absolute;
```
TarasHoliuk: why do you need absolute positioning here? There are easier (and better) ways to center it
Using absolute position just to center element is not good practice, if there are other ways to do it

src/styles/index.scss
```diff
@@ -1,3 +1,65 @@
...
+$transition-duration: 0.5s;
+$hover-transform: -20px;
+$hover-background: #ffbfcb;
+$gap: 1px;
+
 body {
   margin: 0;
 }
+
+.calendar {
```
TarasHoliuk: Consider moving calendar BEM block to a separate file (according to BEM rules)
Same for variables

TarasHoliuk (changes requested):
Deployed version doesn't work.
Try to fix tests (at least in test results there is a visible difference in horizontal borders)

image

VitaliyBondarenko1982 (approved):
Looks good

PR #3672, opened 2024-08-15
finnished the task

stasLiutyww:
[DEMO LINK](https://Sukulynt.github.io/layout_calendar/)
[TEST REPORT LINK](https://Sukulynt.github.io/layout_calendar/report/html_report/)
❗️ Copy this Checklist to the Pull Request description after links, and put - [x] before each point after you checked it.

### DIFF
--- a/src/styles/blocks/reset.scss
+++ b/src/styles/blocks/reset.scss
@@ -1,6 +1,4 @@
 * {
-  margin: 0;
-  padding: 0;
   box-sizing: border-box;
 }
 
--- a/src/styles/blocks/reset.scss
+++ b/src/styles/blocks/reset.scss
@@ -1,0 +1,10 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
--- a/src/styles/blocks/styles.scss
+++ b/src/styles/blocks/styles.scss
@@ -1,0 +1,43 @@
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  width: $calendar-width;
+  gap: $gap-size;
+  align-self: center;
+  padding: $padding-size;
+
+  &__day {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: $day-size;
+    height: $day-size;
+    background-color: $day-background;
+    border: $days-border-size;
+
+    &:hover {
+      background-color: $day-hover-background;
+      transform: translateY(-20px);
+      transition: 0.5s;
+      cursor: pointer;
+    }
+
+    @for $day from 1 through 31 {
+      &:nth-child(#{$day})::before {
+        content: '#{$day}';
+      }
+    }
+  }
+
+  @each $name, $day in $day-of-week {
+    &--start-day-#{$name} &__day:first-child {
+      margin-left: ($day-size + 1) * $day;
+    }
+  }
+
+  @for $day-count from 28 through 31 {
+    &--month-length-#{$day-count} &__day:nth-child(n + #{$day-count + 1}) {
+      display: none;
+    }
+  }
+}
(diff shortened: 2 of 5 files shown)


src/styles/blocks/reset.scss
```diff
@@ -0,0 +1,10 @@
+* {
+  margin: 0;
+  padding: 0;
```
BudnikOleksii: You don't need to set margins and padding for all elements, reset only for tags that actually need this, for example div and span tags don't have any of them and so on

src/styles/index.scss
```diff
@@ -1,3 +1,21 @@
-body {
+@import 'variables/variables';
+@import 'blocks/reset';
+
+* {
   margin: 0;
+  font-family: Arial, Helvetica, sans-serif;
```
BudnikOleksii: And font-family you can set to html or body tags instead

BudnikOleksii (changes requested):
Good work 👍 
Just two minor suggestions regarding `*` selector

src/styles/blocks/reset.scss
stasLiutyww: I think as it is a small project I want to keep it as it wont impact on other things and I expect that I wont add anything more then. And by having margin 0 and padding 0 its easy for me to center elements correctly. So its just my point view.

BudnikOleksii: > I think as it is a small project I want to keep it as it wont impact on other things and I expect that I wont add anything more then. And by having margin 0 and padding 0 its easy for me to center elements correctly. So its just my point view.

You can reset margins and padding only for tags that actually have them, div and span tags don't have any of them and so on

src/styles/index.scss
BudnikOleksii: > And font-family you can set to html or body tags instead

Not fixed, and you use `*` selector twice

BudnikOleksii (changes requested):
Good work 👍 
But a few comments are still not fixed

src/styles/index.scss
```diff
@@ -1,3 +1,16 @@
...
+}
+
 body {
-  margin: 0;
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
```
VolodymyrKirichenko, line 14: move it to `reset.scss` file, where u already use `body` tag

src/styles/index.scss
```diff
@@ -1,3 +1,16 @@
...
+  font-size: $font-size;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
 body {
-  margin: 0;
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
VolodymyrKirichenko, line 13: Note only:
Set standard logic for the tag only if all pages in your project will use similar logic.
It is better to hang the centering on a class, not a tag, because in the future there will be more and more pages.

In this case it works, but keep it in mind for the future

src/styles/blocks/styles.scss
```diff
@@ -0,0 +1,44 @@
...
+    align-items: center;
+    width: $day-size;
+    height: $day-size;
+    background-color: $day-background;
+    border: $days-border-size;
+
+    &:hover {
+      background-color: $day-hover-background;
+      transform: translateY(-20px);
+      transition: 0.5s;
```
VolodymyrKirichenko, line 22: use it on selector, not on `hover` only.
This advice was given to u by a another mentor in the past task, are u forget about it?

src/styles/index.scss
```diff
@@ -1,3 +1,16 @@
+@import 'variables/variables';
+@import 'blocks/reset';
+
+html {
+  font-size: $font-size;
```
VolodymyrKirichenko, line 5: do u need this variable? but why?
u used it only once.

VolodymyrKirichenko (approved):
Good job! 👍 
Please review my comments and correct them to improve your code

PR #3674, opened 2024-08-15
add solution

KrivorotkoYurii:
- [DEMO LINK](https://KrivorotkoYurii.github.io/layout_calendar/)
- [TEST REPORT LINK](https://KrivorotkoYurii.github.io/layout_calendar/report/html_report/)

P.S.: i also added a reset file, file with variables, main.scss file and added path to index.scss file

P.P.S.: After your comment i fixed errors. Visualy it is good but tests go failed. I was  at Q&A session and mentor told me that i need to deploy this version and he will also look if he has time.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -28,7 +28,15 @@
     height: $cell-size;
     width: $cell-size;
     border: $border-property;
-    position: relative;
+
+    // position: relative;
+
+    font-family: Arial, sans-serif;
+    font-size: 30px;
+
+    display: flex;
+    justify-content: center;
+    align-items: center;
 
     transition:
       transform $transition-time,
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+        content: '#{$date}';
+        font-family: Arial, sans-serif;
+        font-size: 30px;
+        line-height: 30px;
+        text-align: center;
+        position: absolute;
+
+        transform: translate(50%, 50%);
+        bottom: 49px;
+        right: 48.5px;
```
BudnikOleksii: - These styles are the same for all elements, so you can set them outside the loop
- To center elements, it's better to use display flex, and avoid using values such as  `48.5`

BudnikOleksii (changes requested):
Good job 👍 
Left a few minor suggestion, check them

Moroz-Dmytro (approved):

PR #3681, opened 2024-08-16
add Calendar

viktoribu:
- [DEMO LINK](https://viktoribu.github.io/layout_calendar/)
- [TEST REPORT LINK](https://viktoribu.github.io/layout_calendar/report/html_report/)
- 

### DIFF
--- a/src/styles/blocks/calendar.scss
+++ b/src/styles/blocks/calendar.scss
@@ -8,8 +8,6 @@
   width: ($amount-of-day * $day-size) + (($amount-of-day - 1) * $gap-between-days) + 20px;
 
   &__day {
-    position: relative;
-
     display: flex;
     justify-content: center;
     align-items: center;
@@ -20,7 +18,8 @@
 
     border: 1px solid $border-color;
 
-    transition: all 0.5s;
+    transition-property: background-color, transform;
+    transition-duration: 0.5s;
 
     &:hover {
       cursor: pointer;
@@ -31,8 +30,6 @@
     @for $i from 1 through 31 {
       &:nth-child(#{$i})::before {
         content: '#{$i}';
-
-        position: absolute;
       }
     }
   }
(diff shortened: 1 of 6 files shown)


viktoribu:
Screenshot 2024-08-17 at 09 52 36

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+    justify-content: center;
+    align-items: center;
+
+    width: $day-size;
+    height: $day-size;
+    background-color: $day-color;
+
+    border: 1px solid $border-color;
+
+    transition: all 0.5s;
```
BudnikOleksii: Specify to which properties should transition apply

src/styles/blocks/calendar.scss
```diff
@@ -0,0 +1,51 @@
...
+      cursor: pointer;
+      background-color: $hover-day-color;
+      transform: translateY(-20px);
+    }
+
+    @for $i from 1 through 31 {
+      &:nth-child(#{$i})::before {
+        content: '#{$i}';
+
+        position: absolute;
```
BudnikOleksii: Do you need to set a position?
Even if you need this move to the selector itself, it's is the same for all of them so you don't need to do it inside loop

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/styles/blocks/calendar.scss
viktoribu: @BudnikOleksii All changes should be animated with the duration of 0.5s

or no?

src/styles/blocks/calendar.scss
viktoribu: thank you for reviewing!
I've just fixed mistakes.

src/styles/blocks/calendar.scss
BudnikOleksii: > @BudnikOleksii All changes should be animated with the duration of 0.5s
> 
> or no?

It's better to avoid using `all` as it can lead to problems with preformance

BudnikOleksii (approved):
👍

PR #3687, opened 2024-08-17
add task solution

v1RnT:
- [DEMO LINK](https://v1rnt.github.io/layout_calendar/)
- [TEST REPORT LINK](https://v1rnt.github.io/layout_calendar/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,89 @@
+// Color variables
+$border-color: #000;
+$day-color: #eee;
+$hover-color: #ffbfcb;
+
+// Size variables
+$cell-size: 100px;
+$padding: 10px;
+$gap: 1px;
+$days-in-a-week: 7;
+$days: (
+  'mon': 0,
+  'tue': 1,
+  'wed': 2,
+  'thu': 3,
+  'fri': 4,
+  'sat': 5,
+  'sun': 6,
+);
+
+// Page styles
+.page {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  margin: 0;
+  height: 100vh;
+
+  font-family: Arial, Helvetica, sans-serif;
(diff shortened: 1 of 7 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,89 @@
+// Color variables
+$border-color: #000;
+$day-color: #eee;
+$hover-color: #ffbfcb;
+
+// Size variables
```
BudnikOleksii: Remove comments
Also you can create a separate file for variables and each BEM block

src/styles/main.scss
```diff
@@ -0,0 +1,89 @@
...
+      transform 0.5s,
+      background-color 0.5s;
+
+    @for $i from 1 through 31 {
+      &:nth-of-type(#{$i})::before {
+        content: '#{$i}';
+        position: absolute;
+        bottom: ($cell-size - 2) / 2;
+        right: ($cell-size - 3) / 2;
+        transform: translate(50%, 50%);
```
BudnikOleksii: - These styles are the same for all of them, so you can move outside the loop.
- It's better to avoid position absolute, to center the element you can use display flex

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/styles/main.scss
v1RnT: I tried doing it with: 
` display: flex;
  align-items: center;
  justify-content: center;` in calendar__day but when i do so, there's a pixel problem because of which i cant pass the test. So i once again used position absolute but moved it out from loop.
image

BudnikOleksii (approved):
👍

PR #3690, opened 2024-08-18
Develop

coldmelnyk:
- [DEMO LINK](https://coldmelnyk.github.io/layout_calendar/)
- [TEST REPORT LINK](https://coldmelnyk.github.io/layout_calendar/report/html_report/)

- [X] Changing 'month-lengh' and 'start-day' modifier in the code element
- [X] Each day has no modifiers, only class (eg. calendar__day)
- [X] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [X] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -22,8 +22,8 @@ body {
     display: none;
     justify-content: center;
     align-items: center;
-    width: 100px;
-    height: 100px;
+    width: $square-size;
+    height: $square-size;
     background-color: #eee;
     border: 1px solid;
     transition-duration: 0.5s;
@@ -44,16 +44,6 @@ body {
     font-size: 30px;
   }
 }
-
-$day-start-styles: (
-  'mon': 0,
-  'tue': 101px,
-  'wed': 202px,
-  'thu': 303px,
-  'fri': 404px,
-  'sat': 505px,
-  'sun': 606px,
-);
 
 @each $day-of-start, $margins in $day-start-styles {
   .calendar--start-day-#{$day-of-start} :first-child {
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,60 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  display: flex;
+  justify-content: center;
+  max-width: 706px;
+}
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 1px;
+
+  &__day {
+    position: relative;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 100px;
+    height: 100px;
+    background-color: #eee;
+    border: 1px solid;
+    transition-duration: .5s;
+
+    &:hover {
+      cursor: pointer;
+      background-color: #FFBFCB;
+      transform: translateY(-20px);
+    }
+  }
+}
+
+@for $number from 1 through 31 {
+  .calendar :nth-child(#{$number})::before {
+    content: '#{$number}';
+    position: absolute;
+    font-family: Arial, Helvetica, sans-serif;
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+  &__day {
+    position: relative;
+    display: none;
+    justify-content: center;
+    align-items: center;
+    width: $square-size;
+    height: $square-size;
+    background-color: #eee;
+    border: 1px solid;
+    transition-duration: 0.5s;
```
BudnikOleksii, line 53: Specify to which properties should transition apply

src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+}
+
+@for $number from 1 through 31 {
+  .calendar :nth-child(#{$number})::before {
+    content: '#{$number}';
+    position: absolute;
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 30px;
```
BudnikOleksii: These styles are the same for all of them, so you can move outside the loop.

src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+    &:hover {
+      cursor: pointer;
+      background-color: #ffbfcb;
+      transform: translateY(-20px);
+    }
+  }
+}
+
+@for $number from 1 through 31 {
+  .calendar :nth-child(#{$number})::before {
```
BudnikOleksii: Use nesting

src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+  .calendar--start-day-#{$day-of-start} :first-child {
+    margin-left: $margins;
+  }
+}
+
+@for $day-show from 29 to 32 {
+  .calendar--month-length-#{$day-show} :nth-child(-n + #{$day-show}) {
+    display: flex;
+  }
+}
```
BudnikOleksii: Why do you do it inside a loop?

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them

src/styles/main.scss
coldmelnyk: Because I do display none for every day and this loop makes elements visible for quantity that sets in month-length modifier 

displays-in-calendar

src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 30px;
```
BudnikOleksii, line 7: - You don't need to specify font for each element, just to html
- And the same with margins/paddings, reset them for tags that actually need it, as div/span/some other tags don't have any of them

src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  align-content: center;
+  max-width: 706px;
+  gap: 1px;
+
+  @for $number from 1 through 31 {
+    & :nth-child(#{$number})::before {
+      content: '#{$number}';
+      position: absolute;
```
BudnikOleksii, line 26: Move position absolute outside the loop, and actually as I see you have centered with display flex, so maybe you don't need to set position for these elements at all?

src/styles/main.scss
BudnikOleksii: > Because I do display none for every day and this loop makes elements visible for quantity that sets in month-length modifier
> 
> displays-in-calendar

Got it

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few suggestions for the future

PR #3760, opened 2024-09-20
add task solution

nadiia21:
- [DEMO LINK](https://nadiia21.github.io/layout_calendar/)
- [TEST REPORT LINK](https://nadiia21.github.io/layout_calendar/report/html_report/)


etojeDenys (changes requested):
you should move the cell on hover
image

src/styles/index.scss
```diff
@@ -40,7 +39,8 @@ body {
     &:hover {
       cursor: pointer;
       background-color: $bg-hover;
-      transition: transform 500ms;
+      transform: translateY(-20px);
+      animation: transform 500ms;
```
etojeDenys, line 43: you should use here the transition instead of the animation

etojeDenys (approved):

PR #3861, opened 2024-10-18
add task solution

RuslanMudryi:


src/styles/index.scss
```diff
@@ -1,3 +1,60 @@
...
+    transform: translate(0, -20px);
+    background-color: $day-color-hover;
+  }
+
+  @for $i from 1 through 31 {
+    &--month-length-#{$i} &__day:nth-child(n + #{$i + 1}) {
+      display: none;
+    }
+
+    @for $o from 1 through $i {
```
BudnikOleksii, line 53: You don't need a nested loop in this case

src/styles/index.scss
```diff
@@ -1,3 +1,60 @@
...
+      &--month-length-#{$i} &__day:nth-child(n + #{$o})::before {
+        content: '#{$o}';
+        margin-left: 1px;
```
BudnikOleksii, line 56: Don't use loops for styles that stay the same

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solutions slightly. Also, please, add a demo link to the PR description so a reviewer can test your solution

PR #4377, opened 2025-04-10
123

BogdanBondar-fw:


src/styles/index.scss
```diff
@@ -1,3 +1,85 @@
...
+
+  @for $i from 1 through 31 {
+    &__day:nth-child(#{$i})::before {
+      content: '#{$i}';
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      font-family: Arial, sans-serif;
+      font-size: 30px;
```
BudnikOleksii, line 76: - Don't use loops for styles that stay the same
- To center an element it's better to use `flex` instead of absolute positioning

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required. Also, add a demo and tests link to the PR description so a reviewer can test your solution

PR #4424, opened 2025-04-15
Solution

NDie1311:
I was advise to remove any modifiers to my individual days but there are none applied.
The modifiers are applied to my calendar--start-day-sun and calendar--month-length--31. Which was defined as requirements in the readme.
Please advise


NDie1311:
The test fails I have done as requested and what the requirements was but still it fails

PR #5096, opened 2025-12-22
add solution

thmoro:


thmoro:
@mateacademy-ai-mentor withou the 'transition: none', the hover test is failling at the BackstopJS, it looks like the screenshot is taken before the transform is completed, how can I fix this?

PR #5367, opened 2026-05-08
add task solution

Isabella-Braga:
- [DEMO LINK](https://Isabella-Braga.github.io/layout_calendar/)
- [TEST REPORT LINK](https://Isabella-Braga.github.io/layout_calendar/report/html_report/)

❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.

- [ X] Changing 'month-lengh' and 'start-day' modifier in the code element
- [ X] Each day has no modifiers, only class (eg. calendar__day)
- [ X] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ X] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)


src/styles/variables.scss
```diff
@@ -0,0 +1,10 @@
+$day-size: 100px;
+$gap-size: 1px;
+$calendar-padding: 10px;
+$days-per-week: 7;
+$start-days: (sun 0, mon 1, tue 2, wed 3, thu 4, fri 5, sat 6);
```
raulriato: You're starting the calendar on `sun` being the first column, but the checklist explicitly says to start on `mon` as the first column:

Image

raulriato (changes requested):
Your code is well structured and works. Just need a tiny fix and you'll be good to go.

src/index.html
```diff
@@ -9,10 +9,42 @@
     <title>Calendar</title>
     <link
       rel="stylesheet"
-      href="styles/index.scss"
+      href="styles/main.scss"
     />
   </head>
   <body>
-    <h1>Calendar</h1>
+    <div class="calendar calendar--start-day-mon calendar--month-length-31">
```
raulriato: That was not the problem. your calendar day start should be sunday. The problem I reported to you was in your `variables`. You're using `sunday` as your first `column` in the calendar there, but it should be `monday`.

src/styles/main.scss
```diff
@@ -0,0 +1,60 @@
+@import './variables';
+@import './index';
+
+.calendar {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $gap-size;
+
+  padding: $calendar-padding;
```
raulriato: This is also a problem. You don't need to have any `padding` on your calendar.

raulriato (changes requested):
You didn't fix the real problem yet. But you're almost there.

src/styles/variables.scss
```diff
@@ -2,9 +2,7 @@ $day-size: 100px;
 $gap-size: 1px;
 $calendar-padding: 10px;
```
raulriato, line 3: You forgot to delete this variable. It's not being used anymore.

raulriato (approved):
I can see you corrected your code, and it's now working, so I'll approve.

Just remember to always run the `npm run deploy` after pushing your code to `Github`, otherwise your deploy will not be up to date with your code and it may cause the mentors to not be able to review it correctly.

PR #5563, opened 2026-08-07
finished solution

hol178g-bit:
- [DEMO LINK](https://hol178g-bit.github.io/layout_calendar/)
- [TEST REPORT LINK](https://hol178g-bit.github.io/layout_calendar/report/html_report/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,10 @@ This is possible because [we use the Parcel library](https://en.parceljs.org/scs
 
 ## Checklist
 
-❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
+❗️ Replace `hol178g-bit` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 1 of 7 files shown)


Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Tablet with `32GB` and the same model but with `128GB` on board are different items - so, you need to handle them separately. 
Now, if I click `Add to Cart` button on the first one, all tablets, including different colors & memory caps are added to Cart automatically
Would you mind to fix it?

Image

Image

Hint: on clicking `color` point or `memory` button, your App should redirect User to the page of the same item, but with chosen color/memory capacity

2.  When User clicks one of proposed in `You may also like` section cards, there should be smooth scroll to the top of the page

3. Sum of the total pre-selected goods items should appear as `Total for ${number} items` - the same as number inside the small red circle in the top right corner of the page

Image

Anton-Kuchmasov (approved):
Good job!

PR #5571, opened 2026-08-13
feat implement calendar layout using SCSS

lucasnunesmarcelino-ctrl:
- [DEMO LINK](https://lucasnunesmarcelino-ctrl/layout_calendar/)
- [TEST REPORT LINK](https://lucasnunesmarcelino-ctrl/layout_calendar/report/html_report/)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,29 @@
+name: Test
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [20.x]
+
+    steps:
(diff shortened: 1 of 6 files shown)


DanielCamposDevX (changes requested):
Good progress — the main calendar structure and SCSS logic are implemented correctly, including BEM classes, day numbering, start-day modifiers, and hover behavior.

Please fix a few things before approval:

Avoid the universal * selector because it is prohibited by our Code Style Rules.
You already created $hover-offset, so please use it in transform instead of hardcoding -20px.
Please make the month-length implementation explicitly cover all required values: 28, 29, 30, and 31.
Fix the Demo and Test Report URLs in the PR description — they should use your GitHub Pages domain (<username>.github.io).

After these changes, the solution should be in good shape for another review.

